PR #133, opened 2015-03-11
Add print this page link to manuals to print full page

russellthorn:
Add print this page link to manuals to print full page even folded
sections on subsection pages.

To make printing easier/more obvious to users who want to print
manuals we've added a "Print this page" to all pages on manuals.
This works for HMRC manuals too.

• added the link to application.html.erb so that it appears on
all pages
• updated styles with input from Rebecca Cottrell on layout/spacing
• added default styles hide element, but included styles to show
element when javascript is enabled
• tested on mobile and IE6, 7, 8 and 9 and works as expected

https://www.pivotaltracker.com/story/show/88435406

manual before -
screen shot 2015-03-11 at 17 09 14
manual after -
screen shot 2015-03-11 at 17 08 38

HMRC before -
screen shot 2015-03-11 at 17 09 34
HMRC after -
screen shot 2015-03-11 at 17 08 12

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -25,7 +25,7 @@ main {
 }
 
 #manuals-frontend {
-  padding-bottom:$gutter*3;
+  padding-bottom:$gutter;
 
   .govuk-beta-label {
     border-bottom: none;
@@ -192,7 +192,7 @@ main {
   }
 
   .manual-body {
-    margin-top: $gutter-half;
+    margin: $gutter-half 0 $gutter*3;
     @include media(tablet){
       margin-top: $gutter;
     }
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/application.scss
```diff
@@ -24,8 +24,17 @@ main {
...
 }
 
+.print-page {
+  display: none;
+  margin-top:$gutter;
+
+  .js-enabled & {
+    display: block;
+  }
+}
```
edds: Is there any reason this isn't nested within the `#manuals-frontend` selector? It would fit  better with the approach the rest of the Sass has taken in this file if it was nested.

russellthorn: There was a reason at first, before I used the .js-enabled & method, but that seems to work fine now. So I've updated this.

PR #134, opened 2015-03-12
Improvements to print stylesheets for manuals

russellthorn:
Improvements made are to improve design and layout of the printed page
and to improve readability.

See screenshots attached for side by side comparisons.
(left image is after, right image is before)

Changes made -
• Hide 'open all' and 'close all' links
• Hide 'Give feedback about this page' link
• clear main content from floating up next to the header metadata
• Remove URLs displaying after header metadata links
• display dt's and dl's inline to match metadata format seen on other
publication formats
• update spacing for h1 in HMRC header to match on screen look closer
• bold h1 for clarity
• bold sub headings for clarity and better scanning

https://www.pivotaltracker.com/story/show/90180042

screen shot 2015-03-12 at 16 40 38
screen shot 2015-03-12 at 16 40 57
screen shot 2015-03-12 at 16 41 19
screen shot 2015-03-12 at 16 41 36
screen shot 2015-03-12 at 16 41 53

### DIFF
--- a/app/assets/stylesheets/_manuals-print.scss
+++ b/app/assets/stylesheets/_manuals-print.scss
@@ -10,57 +10,49 @@ a {
 
 .primary {
   h1 {
-    margin-bottom: 5px;
+    margin: 0 0 10px;
+    padding: 0 0 10px;
+    font-weight: bold;
   }
 
   dl {
     float: left;
     margin-top: 0;
-    width: 25%;
-    dt,
-    dd {
-      display: block;
+    width: 100%;
+    dt {
+      min-width: 100px;
+      float: left;
+      &:after {
+        content: ":";
+      }
     }
     dd {
       font-weight: bold;
       margin-left: 0;
-      & a[href^="http://"]:after,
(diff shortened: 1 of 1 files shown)


evilstreak:
:+1: Looks good, and this PR is great.

In the "Updated" metadata we show "11 February 2015, see all updates" – would it be better to hide that "see all updates" part? That shouldn't block this work being merged, so if we want to do that we could pick it up in another pull request.

russellthorn:
I left that for a few reasons.
• It's a bigger issue than just fixing it for manuals
• There's a hard coded trailing comma after the date which would need hiding if the "see all updates" is removed and there is currently now way of pin pointing that with CSS
• There's no way to currently scope the "see all updates" easily or in a non ugly way with CSS i.e. would require it being wrapped in a <span> or something
• This would need to be updated in the component so that we could hide properly using CSS
• All apps that have similar metadata would need updating or at least need updating to use the component and then the CSS would probably need to live in static

I will write this up as a new piece of work for a later date.

PR #132, opened 2015-03-06
[DISCUSS - DO NOT MERGE] Add page navigation to hmrc manuals

jackscotti:
Adding navigational links to HMRC manuals hasn't been simple and many people had many different ideas on how to approach the issue. 

1 - Get HMRC to provide us previous and next links and simply displaying them within a view in manuals-frontend
2 - Get the content-store to calculate previous and next links and attach them to section schemas
3 - Calculate previous and next links within manuals-frontend:
4 - Doing it within the ManualsController
5 - Doing it within the DocumentPresenter
6 - Create an API app that will live between the content-store and manuals-frontend

I've implemented points 4 and 5.

the last commit addresses 5 the previous ones address 4.

DISCLAIMER: refactoring needed, commit messages re-editing needed

@fatbusinessman  @alicebartlett  @evilstreak

PR #135, opened 2015-03-13
Refactor manuals controller

evilstreak:
Refactor the way manuals and documents are loaded and checked by the manuals controller. Push everything towards smaller methods with a single purpose. I believe these changes make it easier to see how the controller works, and will make it more straightforward to change the way manuals and documents are loaded in the future (eg as part of #132)


jackscotti:
:+1:

PR #136, opened 2015-03-18
Add previous and next links to HMRC manual

jackscotti:
This PR will allow users to navigate through sections within the HMRC manual.

ticket: https://www.pivotaltracker.com/story/show/88435524

all done with the help of the splendid @fatbusinessman, @alicebartlett and @evilstreak

### DIFF
--- a/spec/unit/sibling_decorator_spec.rb
+++ b/spec/unit/sibling_decorator_spec.rb
@@ -1,0 +1,271 @@
+require 'rails_helper'
+
+describe SiblingDecorator do
+
+  let(:document_base_path) {
+    "/guidance/a-manual/child-section"
+  }
+
+  let(:manual_base_path) {
+    "/guidance/a-manual"
+  }
+
+  let(:document_id) {
+    "child-section"
+  }
+
+  let(:document) {
+    double(
+      base_path: document_base_path,
+      details: double(
+        breadcrumbs: [
+          double(
+            section_id: parent_id,
+            base_path: parent_base_path,
+          ),
+        ],
+        section_id: document_id,
+      ),
+    )
+  }
+
+  let(:parent_base_path) {
+    "/guidance/a-manual/parent-section"
+  }
+
+  let(:parent_id) {
+    "parent-section"
+  }
+
+  let(:parent) {
+    double(
+      base_path: parent_base_path,
+      details: double(
+        section_id: parent_id,
+        child_section_groups: child_section_groups,
+      ),
+    )
+  }
+
+  let(:child_section_groups) {
+    []
+  }
+
+  subject(:decorator) {
+    SiblingDecorator.new(
+      document: document,
+      parent: parent,
+    )
+  }
+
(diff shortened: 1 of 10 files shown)


spec/unit/sibling_decorator_spec.rb
```diff
@@ -0,0 +1,270 @@
...
+
+    describe "#next_sibling" do
+      it "should be present" do
+        expect(decorator.next_sibling.section_id).to eq("next-sibling")
+        expect(decorator.next_sibling.base_path).to eq("/guidance/a-manual/next-sibling")
+        expect(decorator.next_sibling.title).to eq("Next sibling title")
+      end
+    end
+
+    context "for a section not recognised by its parent a child" do
```
evilstreak: I think the "a child" at the end is probably a mistake

spec/unit/sibling_decorator_spec.rb
```diff
@@ -0,0 +1,270 @@
...
+        it "should be nil" do
+          expect(decorator.next_sibling).to be_nil
+        end
+      end
+    end
+  end
+
+  context "for a section that is the first child and has a cousin" do
+    let(:child_section_groups) {
+      [ double(
```
evilstreak: For consistency with the rest of this file, this line should be split in two, ie

```ruby
[
  double(
```

evilstreak:
:+1: This looks good. I'll leave someone else who was less involved to review and merge

spec/unit/sibling_decorator_spec.rb
jackscotti: was supposed to be 'as a child', but I can just remove it. Thanks for spotting it!

jackscotti:
@evilstreak addressed all your comments

spec/unit/sibling_decorator_spec.rb
```diff
@@ -0,0 +1,271 @@
...
+
+    describe "#previous_sibling" do
+      it "should be nil" do
+        expect(decorator.previous_sibling).to be_nil
+      end
+    end
+
+    describe "#next_sibling" do
+      it "should be present" do
+        expect(decorator.next_sibling.section_id).to eq("next-sibling")
```
mikejustdoit: Maybe they won't be changing now that you've put them there, but have you considered using variables for 'next sibling' or its associated strings instead of comparing against string literals?

mikejustdoit:
This looks good to me.

Though this is my first encounter with Manuals Frontend and some of its dependencies, so I'm not sure how qualified I am to merge it?

evilstreak:
You have to start somewhere @mikejustdoit!

You are definitely qualified enough to merge this, there's no minimum bar of app experience required. If you don't feel confident enough to merge because the things looks especially involved or are in a language you aren't familiar with, then go for a thumbs up with some notes about why you don't feel confident merging.

jackscotti:
just do it @mikejustdoit !

PR #140, opened 2015-04-13
Add search box to manuals header

alicebartlett:
screen shot 2015-04-13 at 14 59 58

So that users can search within a manual, this commit adds a search box
to the manual header.


app/assets/stylesheets/application.scss
```diff
@@ -104,6 +104,45 @@ main {
...
+
+      button {
+        width: 33px;
+        height: 33px;
+        border: 1px solid #0b0c0c;
+        border-width: 0 0 0 1px;
+        border-left-color: #222;
+        -webkit-border-radius: 0;
+        -moz-border-radius: 0;
+        border-radius: 0;
```
edds: Could you use the mixin for this please to keep the vendor prefixing in one place: https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/_css3.scss#L11-L15

app/presenters/manual_presenter.rb
```diff
@@ -14,6 +14,10 @@ def beta?
     true
   end
 
+  def slug_for_search
+    url[1..-1]
```
jamiecobbett: I'm a little worried that this implementation could easily be broken, especially as there don't appear to be any tests. A comment to document the intent would also help.

Finders include the slug in their content item to enable this kind of integration with search - perhaps we should do that for manuals?

evilstreak: Adding this information into the content item would be a nicer way to do this.

Since updating the publishing tool to include that and republishing all the manuals is somewhat outside the scope of this work, and shouldn't block getting this live, I'm going to merge this now and leave that tidy up for another time.

PR #144, opened 2015-05-07
Update manuals to use govspeak component to update styling

russellthorn:
Manuals wasn't using the govspeak component and therefore was missing
some styles for callouts. So I've updated it to use the component and
deleted the unnecessary govspeak.scss and removed the reference to
that in application.scss

I added in a margin-top overide for .call-to-action paragraphs that
were inheriting it from .collapsible-subsections p's

https://www.pivotaltracker.com/story/show/92362708

before -
screen shot 2015-05-07 at 09 57 41

after -
screen shot 2015-05-07 at 09 57 12


alicebartlett:
OK, I had a little look at this failing test. Using Capybara's `print page.html` I can see that the govspeak component is not rendering html in the page in the test env, instead we have:
```
 <div class="body-content-wrapper">
  <test-govuk-component data-template="govuk_component-govspeak">
    {"content":"On this page:&lt;br&gt;&lt;br&gt;&lt;a href=\"/hmrc-internal-manuals/#{manual_id}/# {section_id}#EIM15010#IDAZR1YH\"&gt;Sections 386-400 ITEPA 2003]&lt;/a&gt;&lt;br&gt;","govspeak":null}
  </test-govuk-component>
 </div>
```
(nb newlines added for legibility)

I'm currently investigating if Slimmer/test can be amended to make this work properly, but if anyone reading this knows the answer and can save me some time, let me know.

jamiecobbett:
@alicebartlett it is intended that it dumps the arguments as JSON rather than doing any processing. See: https://github.com/alphagov/slimmer#testing-shared-components

alicebartlett:
@jamiecobbett Yes! Sorry, I went and read that and then amended the comment but didn't save it. Good work Alice.

So, @russellthorn you should be able to amend the tests as described in https://github.com/alphagov/slimmer#testing-shared-components and it will all work.

dsingleton:
The JSON dump is so the testing app can parse the inputs to a component. This need to be done manually at the moment (parsing out the JSON and evaluating it). I think there are some examples, but i can't find one right now :(

Longer term the plan is to add some test helpers to Slimmer, so a test doesn't need to know about how how to select or parse `test-govuk-component`. Hopefully this'll get easier as we start standardising on a smaller set of test frameworks/versions.

alicebartlett:
Thanks for fixing that test, happy to merge.

PR #145, opened 2015-05-12
Contract tests for HMRC manuals

jamiecobbett:


spec/features/viewing_a_manual_spec.rb
```diff
@@ -181,12 +181,12 @@
   describe "support for example manual sections from govuk-content-schemas" do
```
heathd: I feel like it would be preferable to put the contract tests in a separate file from the view tests. The view tests use fixtures internal to this repo (as they should) wheras the contract tests explicitly use the examples from `govuk-content-schemas` to test integration. Having them in the same file blurs this distinction and makes it easier for a future reader to overlook this important difference in purpose between the two sets of tests.

I'd suggest moving all of the tests which rely on examples govuk-content-schemas into a file called something like `govuk_content_schema_contract_spec.rb` or `govuk_content_schema_integration_spec.rb` or maybe in a separate folder e.g. `spec/contracts/govuk_content_schemas_spec.rb`

jamiecobbett: Fair enough - added.

PR #151, opened 2015-07-15
Replace andSelf with addBack

edds:
andSelf is being deprecated in future versions of jQuery, swap to using
addBack instead which is the future proof way of achieving the same
effect.


edds:
Ugh, I was testing against a new jQuery. Looks like we can't just do this like this. Closing while I find a new solution.

PR #155, opened 2015-08-20
Remove employment income manual code

tommyp:
This manual was used as a placeholder to demonstrate Manuals Frontend
before we had any content. The placeholder is no longer being used so
this commit removes the `before_filter`, the tests and a rake task which
publishes a route for the manual.


jamiecobbett:
Oh, the memories :crying_cat_face:

spec/features/viewing_a_manual_spec.rb
```diff
@@ -128,23 +128,6 @@
...
-  scenario "visiting a section in the obsolete prototype HMRC manual" do
-    content_store_does_not_have_item('/hmrc-internal-manuals/employment-income-manual')
-    content_store_does_not_have_item('/hmrc-internal-manuals/employment-income-manual/abc123')
-
-    visit_hmrc_manual_section "employment-income-manual", "abc123"
-    expect_manual_title_to_be("This prototype of the Employment Income Manual is no longer available")
-    expect(page.status_code).to eq(200)
-  end
-
   context "prototype HMRC manual has now been published for real" do
```
jamiecobbett: It probably makes sense to get rid of this test too, it doesn't make sense with the fallback code gone.

tommyp: Oh yeah! Deleted it. I forgot to read any further down the file...

lib/tasks/publishing_api.rake
```diff
@@ -41,16 +41,6 @@ namespace :publishing_api do
...
     logger = Logger.new(STDOUT)
 
-    special_route_publisher = GdsApi::PublishingApi::SpecialRoutePublisher.new(logger: logger)
-    special_route_publisher.publish(
-      content_id: "8a97dc8e-d48f-48ad-804e-862d327e1fc1",
-      title: "Prototype of the Employment Income Manual (defunct)",
-      base_path: "/guidance/employment-income-manual",
-      type: "prefix",
-      publishing_app: 'manuals-frontend',
-      rendering_app: 'manuals-frontend'
```
jennyd: We should make sure that this route is properly withdrawn too.

jennyd:
The [view for the manual](https://github.com/alphagov/manuals-frontend/blob/d3b0f8ff97edaea3fb29a4c23978f44399b3f512/app/views/manuals/employment_income_manual.html.erb) won't be used any more, so we should remove that too.

tommyp:
I've made use of the new `SpecialRouteUnpublisher` that I've added in https://github.com/alphagov/gds-api-adapters/pull/350 in order to remove the route for this. That PR needs merged and `gds-api-adapters` needs bumped in order for this to go green.

tommyp:
I've updated this PR to publish a gone content item to the Publishing API. This code can be removed after this is merged and deployed.

PR #156, opened 2015-08-24
For discussion: Disable caching for content-store

bradleywright:
To discuss: is this a pattern we want to generally apply? In `government-frontend`, we [explicitly pass the `cache-control` headers downstream](https://github.com/alphagov/government-frontend/blob/master/app/controllers/content_items_controller.rb#L37-L41), relying on those clients to cache the response.

We shouldn't cache content-store items in the controller requests, as this results in an inconsistency between what is published in the API and what is available to clients hitting the controllers directly.

Further to this, Rails's built-in caching is opaque to external systems, and requires a hard-restart before being cleared. This gives us no options in the future to easily propagate content-store changes to downstream clients.

### DIFF
--- a/app/controllers/manuals_controller.rb
+++ b/app/controllers/manuals_controller.rb
@@ -40,7 +40,7 @@ private
   end
 
   def manual
-    content_store.content_item(manual_base_path)
+    content_store(disable_cache: true).content_item(manual_base_path)
   end
 
   def load_manual
--- a/app/repositories/document_repository.rb
+++ b/app/repositories/document_repository.rb
@@ -20,7 +20,7 @@ private
   end
 
   def content_store
-    GdsApi::ContentStore.new(Plek.new.find("content-store"))
+    GdsApi::ContentStore.new(Plek.new.find("content-store"), disable_cache: true)
   end
 
   def extract_parent_base_path_from_document(document)


boffbowsh:
I'd agree with this. Are you saying we should also add the cache headers in the same way as in `government-frontend` too? That also makes sense to me, are there reasons this doesn't currently happen?

bradleywright:
> Are you saying we should also add the cache headers in the same way as in government-frontend too? That also makes sense to me, are there reasons this doesn't currently happen?

I think we should. It doesn't currently happen because that pattern came along after the manuals frontend code was written.

boffbowsh:
:+1: for this + passing the cache headers

jennyd:
:+1: to not caching the content-store response in manuals-frontend, for the reasons already mentioned above. I've found the caching in API adapters confusing in other contexts recently as well.

Re setting cache headers based on data in the content item, that seems like a good pattern in general, but I'm not sure where that data is coming from - does it rely on the publishing app sending it to publishing-api? The original reasoning for passing on the headers in government-frontend was to [prevent caching of access-limited content](https://github.com/alphagov/government-frontend/pull/50), which we don't have for manuals.

bradleywright:
@jennyd access limited items automatically set no-cache headers in content-store:

https://github.com/alphagov/content-store/blob/master/app/controllers/content_items_controller.rb#L11

bradleywright:
Oh sorry, misunderstood the question.

So, eventually things like scheduling, access limiting etc. will be a centralised function, and it will depend on the state transition and content type what appropriate headers are. Frontend apps should just do a transparent pass-through of that information, as publishing apps + other internal state can best determine what those headers should be.

bradleywright:
And on that, the default cache time is [30.minutes](https://github.com/alphagov/content-store/blob/master/config/application.rb#L44).

bradleywright:
I'll close this now and see if I can get to adding cache pass-through this week.

PR #159, opened 2015-09-16
Don't strip leading slashes from the manual URL anymore

issyl0:
- This was used when searching within a manual as we stripped leading
  slashes on all other occurrences of the manual slug. We shouldn't do
  this anymore, because it breaks search within a manual in another way
  (which required a temporary fix in frontend).
- Trello:
  https://trello.com/c/KAR7YtA7/119-searching-within-a-manual-uses-the-wrong-layout.

### DIFF
--- a/app/presenters/manual_presenter.rb
+++ b/app/presenters/manual_presenter.rb
@@ -14,10 +14,6 @@ class ManualPresenter
     true
   end
 
-  def slug_for_search
-    url[1..-1]
-  end
-
   def beta_message
     if hmrc?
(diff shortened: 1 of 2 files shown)


issyl0:
Closing this until we're ready to deploy specialist-publisher, this change and hmrc-manuals-api close together to avoid more than momentary brokenness for searching within a manual.

issyl0:
Now that https://github.com/alphagov/hmrc-manuals-api/pull/109 is in review, and https://github.com/alphagov/specialist-publisher/pull/564 is merged, reopening this. These three things will need coordinated deployment.

PR #162, opened 2015-10-01
410 on updates with withdrawn manual

tommyp:
## The Problem

`/updates` does not have its own content item, so when withdrawing a Manual a `gone` content item is published for the Manual and each of its sections but not `#{manual_slug}/updates`. When a User views `/updates` on a withdrawn Manual, Manuals Frontend will try to render the page with the `gone` content item that it has for the withdrawn Manual.

When that happens, the App will 500 and this will fall over to the cache, which has the `/updates` page cached and displays it.

## The solution

Don't return a `content_item` for the Manual if it has a format of `gone`, allowing the App to 410 as would be expected.

This PR does just that and adds a failing test for this edgecase.

### DIFF
--- a/app/controllers/manuals_controller.rb
+++ b/app/controllers/manuals_controller.rb
@@ -21,7 +21,11 @@ class ManualsController < ApplicationController
 private
 
   def ensure_manual_is_found
-    error_not_found unless manual
+    if manual.nil?
+      error_not_found
+    elsif manual.format == "gone"
+      error_gone
+    end
   end
 
   def ensure_document_is_found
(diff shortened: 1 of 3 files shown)


mikejustdoit:
Well spotted and nice, simple solution :+1:

jamiecobbett:
I may be missing something, but the gone item should include a route for the /updates path.

tommyp:
If what you're suggesting is adding `/updates` to the routes for the gone content item for a Manual there is an issue with this. I've chatted to Performance Platform and doing that will let the router handle this, but when you publish the Manual again it won't remove the `/updates` route from the Router and fall back to letting Manuals Frontend handle this as it does now. The longer term solution is that `/updates` is its own Content Item. We can add this of the list of things to fix with Specialist Publisher.

PR #164, opened 2015-10-06
Add example URL to README

dsingleton:
Handy for quickly getting an example of a page served by the app.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,11 @@ Manuals Frontend
 
 Front-end app for the manuals format on GOV.UK
 
+## Live examples
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,6 +5,10 @@ Manuals Frontend
 
 Front-end app for the manuals format on GOV.UK
 
+## Live examples
+
+- [gov.uk/guidance/content-design](https://www.gov.uk/guidance/content-design)
```
jamiecobbett: I think its worth adding an HMRC manual example as they do differ in several ways: https://www.gov.uk/hmrc-internal-manuals/vat-government-and-public-bodies

dsingleton: Done.

PR #165, opened 2015-10-28
Fix internal anchor links

davidbasalla:
For https://trello.com/c/E6u89QxG/132-fix-manuals-linking-issue

Fixes the issue of clicking on a link (that links to another section within the same manual) not opening the appropriate section. The fix itself is just changing one line, but in order to write test for it, the code was refactored.


robinwhittleton:
All looks reasonable, my only nit is that you’re using `==` and `!=` instead of `===` and `!==`. It’s good practice to try and avoid implicit type conversions where possible.

davidbasalla:
Comments addressed and rebased into respective commit

PR #166, opened 2015-10-30
Fix multiple creation of close spans

davidbasalla:
This fixes an issue for when a Collapsible would be opened by clicking on a link to it, a new Collapsible was created for it. In practice this meant that a collapsible could have an increasing number of clickable spans, which lead to unexpected such as a click not appearing to open a span (while actually the click was triggering the toggle open action twice, thereby opening and immediately closing the Collapsible). As part of the fix, some functionality was extracted to separate methods and respective Jasmine tests have been added.


davidbasalla:
Comments about using jQuery's `.each` over `Object.keys` addressed and merged into commits

robinwhittleton:
LGTM 👍

PR #168, opened 2016-01-13
Remove beta_message from ManualPresenter

h-lame:
We no longer want to provide a special message in the beta label for HMRC
content as the only manuals of theirs published here are not available in
their library.

For: https://trello.com/c/ZSSTBlqM/245-change-the-beta-label-on-hmrc-manuals

### DIFF
--- a/app/presenters/manual_presenter.rb
+++ b/app/presenters/manual_presenter.rb
@@ -14,12 +14,6 @@ class ManualPresenter
     true
   end
 
-  def beta_message
-    if hmrc?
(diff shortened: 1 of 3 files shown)


h-lame:
The truth of the description is dependant on us running the rake task from https://trello.com/c/XNwq7LPe/244-remove-two-hmrc-manuals-from-gov-uk to remove the two manuals that are available in the library.  We'll do that at the same time we deploy this change.

davidbasalla:
Looks good!

PR #169, opened 2016-01-13
Update manual presenter to prefer links for organisation data

mobaig:
We're currently updating apps to support a new tagging architecture.
Part of this work includes changes to hmrc-manuals-api such that it sets
the organisation data in the links hash rather than the details hash.

This change to manuals-frontend ensures organisation data in the links
field of the manual payload is used in preference to data in details.
This will eventually be updated to drop usage of details altogether.

Trello: https://trello.com/c/6Owhx9Vn

### DIFF
--- a/app/presenters/manual_presenter.rb
+++ b/app/presenters/manual_presenter.rb
@@ -33,7 +33,7 @@ class ManualPresenter
   end
 
   def organisations
-    @manual.details.organisations || []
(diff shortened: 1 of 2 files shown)


mobaig:
@h-lame not going to be an issue in this app but it does highlight that these manual helpers could possibly do with a bit of a rethink so that they make use of the govuk-content-schemas that we have available. Will close this PR for now while I rework this a bit.

PR #172, opened 2016-02-11
Add HTTP caching

davidbasalla:
For https://trello.com/c/pl3zZeWr/257-add-caching-to-manuals-frontend

Manuals-frontend previously did not set any caching, so Rails by default prevents any caching (setting `Cache-Control: max-age=0, private, must-revalidate`).

This implementation is inspired by [`government-frontend`'s implementation of the `set_expiry` method](https://github.com/alphagov/government-frontend/commit/fe66f96c238be1c751c49f54871e2ffec93edcc6), setting cache headers based on the content items that are being presented.

### DIFF
--- a/app/controllers/manuals_controller.rb
+++ b/app/controllers/manuals_controller.rb
@@ -9,9 +9,11 @@ class ManualsController < ApplicationController
   before_action :prevent_robots_from_indexing_hmrc_manuals
 
   def index
+    set_expiry(manual)
   end
 
   def show
+    set_expiry(document)
     @document = DocumentPresenter.new(document, @manual)
   end
 
@@ -41,7 +43,7 @@ private
   end
 
   def manual
-    content_store.content_item(manual_base_path)
+    @_manual ||= content_store.content_item(manual_base_path)
   end
 
   def load_manual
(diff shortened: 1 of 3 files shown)


jamiecobbett:
Might be worth looking at government-frontend's implementation as I think that's quite full-featured: https://github.com/alphagov/government-frontend/blob/master/app/controllers/content_items_controller.rb#L35

h-lame:
The commit message has references to commit-shas from other projects but aren't scoped to those projects so a future-developer might assume they are commit-shas in this project and get confused when they can't find them.  Probably safest to turn them into github urls.

bradwright:
I'd echo @jamiecobbett's comment: the implementation in `government-frontend` is intended to be the future design of frontend caching as it puts the caching headers in the hands of the content items rather than arbitrarily picking a value. It also means we can safely create layers of caching.

davidbasalla:
:+1: thanks for the comments, will update

jamiecobbett:
> the implementation in government-frontend is intended to be the future design of frontend caching as it puts the caching headers in the hands of the content items rather than arbitrarily picking a value. It also means we can safely create layers of caching.

Sounds like we're going to want to share some code between apps.

davidbasalla:
Have updated and rebased. I've altered the method to pass a content item as a parameter (to support both `document` and `manual` content-items), but the functionality remains the same.

Let me know if you spot anything that needs fixing!

h-lame:
Looks good.  Let's try it on integration.

PR #174, opened 2016-05-25
Remove invalid breadcrumbs aria role

fofr:
`role="breadcrumbs"` is not a thing.
https://www.w3.org/TR/wai-aria/roles#role_definitions

We previously removed from the component: https://github.com/alphagov/static/pull/737
Was added in: https://github.com/alphagov/manuals-frontend/pull/4


nickcolley:
LGTM 👍

PR #180, opened 2016-09-15
Add meta descriptions

rubenarakelyan:
The meta description tag is used for the index page of a manual but none of the sub-pages. This commit adds meta description tags to all pages, the contents of which are set to the description associated with the manual. This adds descriptions to external search engine results when the pages are re-indexed.

Trello: https://trello.com/c/bbAwFAQ1/108-add-meta-descriptions-to-pages-missing-them

### DIFF
--- a/app/views/manuals/updates.html.erb
+++ b/app/views/manuals/updates.html.erb
@@ -1,3 +1,7 @@
+<% content_for :extra_headers do %>
+  <meta name='description' content="List of updates to '<%= @manual.full_title %>'." />
+<% end %>
+
 <% content_for :title, 'Updates - ' + @manual.full_title %>
 <%= render partial: "header" %>
 <div class='manual-body' id="content">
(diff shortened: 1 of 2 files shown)


app/views/manuals/updates.html.erb
```diff
@@ -1,3 +1,9 @@
+<% content_for :extra_headers do %>
```
tijmenb: Perhaps this should not be the same as the manual page. Perhaps something like this?

> Overview of changes for "Content design: planning, writing and managing content"

Maybe check with a content person.

rubenarakelyan: After consultation with Alan Madrell, I've gone for "List of updates to '[name of manual].'

tijmenb: Great!

app/views/manuals/updates.html.erb
```diff
@@ -1,3 +1,9 @@
...
+  <% if @manual.summary.present? %>
```
tijmenb: You don't need this conditional anymore I reckon.

rubenarakelyan: Yarp, removed!

PR #189, opened 2016-12-07
Give keyboard access to collapsible headings

fofr:
Quick fix, doesn't fix all accessibility concerns with these toggles; eg ARIA states, odd tabbing behaviour on visually hidden links (patched but not fixed), etc.

The current heading toggles can’t be opened with a keyboard. By dynamically adding links inside the headings we get all the correct tabindex and focus behaviour.

https://www.gov.uk/guidance/content-design/what-is-content-design

* Use the ID of the heading as the href target
* Another JS listener listens for clicks on links. This has been reduced in scope to body links (for expanding/closing footnotes).
* Show section when focussing on hidden links

Because content is visually hidden, the links within the content blocks can still be tabbed too. Rather than tabbing to something you can’t see, which feels broken, if the section isn’t visible, automatically open the section. (Sometimes the browser doesn't scroll to the focused link, sometimes it does.)

tabbing-manuals

@nickcolley @cfq @aduggin

### DIFF
--- a/app/assets/javascripts/collapsible.js
+++ b/app/assets/javascripts/collapsible.js
@@ -7,7 +7,15 @@
     this.$clickTarget = this.$section.find('.js-subsection-title');
     this.$clickTarget.on('click', this.toggle.bind(this));
     this.addToggle();
+
+    this.$section.on('focus', '.js-subsection-body a', this.showSectionWhenLinkFocused.bind(this));
   }
+
+  Collapsible.prototype.showSectionWhenLinkFocused = function() {
+    if (this.$section.is('.closed')) {
+      this.$section.toggleClass('closed');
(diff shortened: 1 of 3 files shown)


cfq:
It would be great to come as close as possible to the example implemented here: http://heydonworks.com/practical_aria_examples/#progressive-collapsibles

I would also be good to reimplement the expand/collapse buttons as <button>s with text inside them rather than Λ and V icons only.

fofr:
@cfq Yep. Those are good points and the example is useful. Is this a good first step towards that by fixing the most egregious problem?

tvararu:
Gave this a play locally, I think this is a solid improvement for UX. Going to give the code a look.

PR #190, opened 2016-12-13
Wrap siblings method in guard conditional

alexmuller:
The content store contains some manuals child pages where the parent doesn't exist. This means that `parent` is nil so `parent.details` causes an error.

This change will allow the child page to render even when it doesn't have a parent.

### DIFF
--- a/app/presenters/sibling_decorator.rb
+++ b/app/presenters/sibling_decorator.rb
@@ -21,9 +21,11 @@ private
   end
 
   def siblings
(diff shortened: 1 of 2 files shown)


jennyd:
We've been trying to get HMRC to fix the data behind these errors for a long time - long enough that I think it's worth doing this. However we would still want to know where we have inconsistent data so that we can feed that back to HMRC, and so that we're aware of how many broken links there will be on the pages (for example in breadcrumbs). We could log these cases - what do you think @sihugh?

PR #195, opened 2017-02-08
Show first published at

h-lame:
As a follow up to the UI work in manuals publisher in https://github.com/alphagov/manuals-publisher/pull/799 we want to expose the `first_published_at` of the manual.  This brings it more in line with other document frontends.

In passing we also change to use the govuk-component for title and metadata, but this does introduce changes to the display of the manual (but it might mean it's more consistent with other gov.uk styles so...).

### Before

the highway code - guidance - gov uk - before

### After

the highway code - guidance - gov uk


fofr:
We opted against using the title component on HTML publications (which also uses a blue bounding box and white text). This was to avoid overriding component styles (eg title colour) and to have nicer margins above the title (the margin looks slightly too large on this PR).

I think this is something @nickcolley was questioning.

app/views/manuals/_header.html.erb
```diff
@@ -7,15 +7,20 @@
...
-        <dt>Updated</dt>
-        <dd><%= marked_up_date(@manual.updated_at) %>, <%= link_to 'see all updates', @manual.updates_url %></dd>
-      </dl>
-    <% end %>
+    <div id="manual-title"><%= render partial: 'govuk_component/title', locals: {
+      title: @manual.title,
+      average_title_length: "long"
+    } %></div>
+    <%= render partial: 'govuk_component/metadata', locals: {
+      from: @manual.organisations.map { |org|
```
fofr: Could some of this logic move into the presenter?

In government-frontend the parameters for the component are managed in the presenter, keeping the template simple and re-usable: eg https://github.com/alphagov/government-frontend/blob/master/app/presenters/metadata.rb#L5 https://github.com/alphagov/government-frontend/blob/master/app/views/content_items/detailed_guide.html.erb#L23

:100: for switching to component

h-lame: Makes sense, will do. I think specialist-frontend does the same thing too.

spec/support/app_helpers.rb
```diff
@@ -1,6 +1,9 @@
 module AppHelpers
   def view_manual_change_notes
-    click_on "see all updates"
+    within('header test-govuk-component[data-template="govuk_component-metadata"]') do
```
fofr: Slimmer provides some test helpers for working with components:
https://github.com/alphagov/slimmer/blob/868095a99fdf559cdad80b9102d882322214bc60/lib/slimmer/test_helpers/shared_templates.rb#L12-L15

Example usage:
https://github.com/alphagov/government-frontend/blob/master/test/test_helper.rb#L33
https://github.com/alphagov/government-frontend/blob/master/test/test_helper.rb#L57-L61

h-lame:
Thanks for looking @fofr. I've fixed up the tests using the helpers you pointed out, they're much nicer now thanks!  

I've somewhat speculatively moved the existing presenters to the models folder to allow me to introduce a "true" presenter for the manuals to encapsulate the work needed to create the args for the metadata component.  Let me know what you think, it's a bunch of work that could be achieved with a traditional helper method, but having it live on the object felt nicer to me.

I've left the title component for now, pending @nickcolley's input.

fofr:
0cb44ac moves this app away from _similar_ behaviour in other frontend applications (namely government-frontend). Is it possible that this refactor can be separated from this PR into a new one and discussed there?

fofr:
New tests look good. :+1:

Not sure how I feel about the presenters move – I don't think I understand enough of the application to give a proper review.

I think the reason you gave about the `aria` attributes being on the wrong element is reason not to use the component.

h-lame:
Thanks @fofr - I've dropped the commit that introduces the component for the title.  I could carve the last 3 commits off into their own PR for discussion with others if you felt more comfortable with that?

h-lame:
Or get another viewpoint on this PR for those 3 commits.

davidbasalla:
The last 3 commits look good to me! It seems to follow the same pattern as specialist frontend or multipage frontend which uses a 'models' to represent the data from content store as well as presenters... 👍

fofr:
> The last 3 commits look good to me! It seems to follow the same pattern as specialist frontend or multipage frontend which uses a 'models' to represent the data from content store as well as presenters... 👍 
We will be removing both of those applications in favour of government-frontend.

PR #196, opened 2017-02-08
Add lint

rubenarakelyan:
Add linting to manuals-frontend, and fix issues that it raises.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -42,6 +42,7 @@ GEM
       builder
       multi_json
     arel (6.0.3)
+    ast (2.3.0)
     better_errors (1.1.0)
       coderay (>= 1.0.0)
       erubis (>= 2.6.6)
@@ -80,6 +81,9 @@ GEM
       activesupport (>= 4.1.0)
     govuk-content-schema-test-helpers (1.3.0)
       json-schema (~> 2.5.1)
+    govuk-lint (1.2.1)
+      rubocop (~> 0.39.0)
+      scss_lint
     govuk_frontend_toolkit (1.2.0)
       rails (>= 3.1.0)
       sass (>= 3.2.0)
@@ -121,6 +125,8 @@ GEM
       mini_portile2 (~> 2.1.0)
       pkg-config (~> 1.1.7)
     null_logger (0.0.1)
+    parser (2.4.0.0)
+      ast (~> 2.2)
     phantomjs (1.9.7.0)
     pkg-config (1.1.7)
     plek (1.11.0)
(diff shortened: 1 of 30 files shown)


fofr:
Can we enable Sass linting too? (It comes with govuk-lint)

Gemfile.lock
```diff
@@ -257,3 +276,6 @@ DEPENDENCIES
   uglifier (>= 1.3.0)
   unicorn (= 4.8.2)
   webmock (~> 2.1.0)
+
+BUNDLED WITH
+   1.12.5
```
h-lame: Ruby 2.3.1 which this project uses is pinned to bundler `1.11.2` (see https://github.com/alphagov/govuk-puppet/blob/9a17813a00e18ad6c22db48ceda24d233d404b24/modules/govuk_rbenv/manifests/all.pp#L56) so this line shouldn't read `1.12.5`.

h-lame:
:clap: for intent! It doesn't seem to pass on CI though, more linting errors to fix?  I'm not sure if we're still doing this, but in other projects I think we were setting up lint to run slightly differently on CI too - in `--diff` mode (and with some extra --format options for output) so we're only checking for lint problems in changed files (useful for when we bump linter and get new rules, or we add linter and don't want to add a big "fix all the things" noisy commit).

Gemfile.lock
rubenarakelyan: Fixed, didn't know about this!

rubenarakelyan:
@fofr Yup, added sass linting
@h-lame CI showed some more linting errors that I fixed. I have changed the rake talk to not run by default, and the Jenkinsfile now runs both linters with the relevant flags set, whereas the rake task runs across the whole codebase.

I have made a load of sass fixes to the stylesheets. I have checked these visually and they look fine but a second eye over it would be great.

app/assets/stylesheets/_reset.scss
```diff
@@ -1,6 +1,8 @@
-// flatten all browser defaults and styles inherited from gov.uk header
-
-/* Borrowed from http://meyerweb.com/eric/tools/css/reset/ */
+// Flatten all browser defaults and styles inherited from gov.uk header
```
fofr: It's weird we have a custom reset in here.

app/assets/stylesheets/application.scss
```diff
@@ -8,8 +8,12 @@
 
 @import "reset";
 
+// some colours used below
+$manual-search-button-colour: #0b0c0c;
```
fofr: I think this can be set to `$black`:
https://github.com/alphagov/govuk_frontend_toolkit/blob/89168934d1455754f455c398ba5c350a484d9a2a/stylesheets/colours/_palette.scss#L50

(I like that the button variable has been extracted out as a concept)

fofr:
Really happy to see the Sass linting fixes :)

PR #198, opened 2017-02-22
Navigation ab test

carvil:
This PR allows us to A/B test the new navigation for education-specific manuals.

There are 3 conditions to meet in order to see the new navigation:
- the manual is tagged to a taxon (currently, only education taxons are available);
- There is an environment variable set to `yes` (ENABLE_NEW_NAVIGATION) - this is try in both development and integration;
- The user is requesting the `B` variant of the A/B test, which is possible to do via the GOV.UK Chrome extension.

When those conditions are met, we:
- add the taxonomy breadcrumbs to the top of the page, deleting the existing breadcrumbs from the page;
- for manual sections, we append the section to the breadcrumb manually;
- we include the beta label, which indicates the new navigation is still in beta.

From this PR, there are 2 things left to do later (comments included in the code as well):
- migrate the `with_b_variant` helper method to `govuk_ab_testing`;
- make sure manual sections are tagged with the taxon of their manual, so that we don't need to "hack" the breadcrumbs that are generated from `govuk_navigation_helpers`.

Note: the new navigation will mostly make space for a side navigation, moving the main content to `column-two-thirds`. However, in this case, @alextea suggested we keep manuals full with. We can always introduce that change if we decide it's best to use it.

I will make sure there are follow up Trello cards for these 2 changes. The tagging of manual sections is not a trivial change to make.

This change relates to the following Trello card:
https://trello.com/c/V6PSacVC/336-changes-to-education-related-content-pages-as-part-of-the-new-navigation-in-manuals-frontend

Similar PRs that introduce A/B testing in other frontend apps:
- https://github.com/alphagov/frontend/pull/1109
- https://github.com/alphagov/smart-answers/pull/2899

Here is how a manual page looks like under the B variant:

screen shot 2017-02-22 at 00 07 01

Here is a manual section page:

screen shot 2017-02-22 at 00 06 33

And here is a page containing manual updates:

screen shot 2017-02-22 at 09 45 23


app/views/manuals/_header.html.erb
```diff
@@ -28,16 +28,16 @@
...
+  <% if request.path == presented_manual.url %>
     <li class="no-separator">Contents</li>
   <% else %>
-    <li><%= link_to 'Contents', manual.url %></li>
+    <li><%= link_to 'Contents', presented_manual.url %></li>
   <% end %>
 
-  <% if @document %>
-    <% @document.breadcrumbs.each do | crumb | %>
+  <% if defined?(document) %>
```
h-lame: Minor and it's up to you if you want to change, but I've always preferred `local_assigns.has_key?  :document` to `defined?(document)` for these cases.

app/controllers/manuals_controller.rb
```diff
@@ -5,33 +5,49 @@ class ManualsController < ApplicationController
...
   end
 
   def show
-    set_expiry(document)
-    @document = Document.new(document, @manual)
+    set_expiry(document_from_repository)
+    document = Document.new(document_from_repository, manual)
+
+    render :show, locals: {
+      presented_manual: presented_manual,
```
h-lame: I like this, making the dependencies to the view explicit is :+1:.  I wonder though if we could call this param `manual` instead though?  It being `presented_manual` doesn't feel that useful in the context of the view (there's no other manual variable to compare it against, and the view doesn't need to know that this is a presenter based manual instance).  If we call it `manual` we remove a bunch of the changes in the view files for this commit.

app/controllers/manuals_controller.rb
```diff
@@ -5,33 +5,49 @@ class ManualsController < ApplicationController
...
+    render :updates, locals: {
+      presented_manual: presented_manual
+    }
   end
 
 private
 
+  def presented_manual
+    @presented_manual ||=
+      ManualPresenter.new(manual: manual, view_context: view_context)
```
h-lame: Oh, yes, this is good.  I forgot that the controller has a `view_context` that we can use instead of relying on `self` in the templates.  :+1:

app/views/manuals/_header.html.erb
```diff
@@ -1,19 +1,19 @@
...
 <% end %>
-<header aria-labelledby="manual-title" class="<%= manual.hmrc? ? 'hmrc' : nil %>">
+<header aria-labelledby="manual-title" class="<%= presented_manual.hmrc? ? 'hmrc' : nil %>">
   <div class='primary'>
-    <% if manual.hmrc? %>
+    <% if presented_manual.hmrc? %>
       <span class='manual-type'>HMRC internal manual</span>
     <% end %>
 
-    <h1 id="manual-title"><%= @manual.title %></h1>
```
h-lame: Oops!  Leaving this as `@manual` is my fault!  Glad you spotted it.

app/views/manuals/_content_for_manual.html.erb
```diff
@@ -0,0 +1,7 @@
+<% content_for :extra_headers do %>
```
h-lame: I don't know about the name for this partial, content_for_manual makes it sound like it's writing out the content of the manual.  I get that content_for is a rails method and thats the only thing we do in this partial, but it doesn't really tell me what this partial is really for - should all content_for calls go in here?  That seems somewhat arbitrary.

app/views/manuals/_content_for_manual.html.erb
```diff
@@ -0,0 +1,7 @@
...
+  <% if presented_manual.summary.present? %>
+    <meta name='description' content='<%= presented_manual.summary %>' />
+  <% end %>
```
h-lame: I wonder if the arguments to this should be `title` and `description` rather than `title` and `presented_manual`.  That way we could re-use it in the updates view.

app/views/manuals/index.html.erb
```diff
@@ -1,36 +1,13 @@
...
-          </li>
-        <% end %>
-        </ol>
-      <% end %>
-    <% end %>
-  </div>
-</article>
+<%= render 'header', presented_manual: presented_manual %>
+<%= render 'manual_breadcrumbs', presented_manual: presented_manual %>
+<%= render 'manual', presented_manual: presented_manual %>
```
h-lame: I don't know that we gain much by extracting the `manual`, `manual_section`, and `manual_updates` partials here - they're not reused elsewhere, and rendering a partial isn't 0-cost in terms of performance (it's tiny, but to be avoided if there's no gain).  I think we can leave their contents inside `index`, `show`, and `updates` respectively without losing anything.

app/views/manuals/updates.html.erb
```diff
@@ -1,40 +1,15 @@
...
-        </div>
-      </div>
-    <% end %>
-    </div>
-  </article>
-</div>
+<%= render 'header', presented_manual: presented_manual %>
+<% if presented_manual.beta? %>
+  <%= render partial: 'govuk_component/beta_label' %>
+<% end %>
```
h-lame: This should go above the header partial, no?  It might be a good idea to put the beta-label bit back into the header partial.  Arguably while the beta-label is not literally inside the `<header>` element, it's part of the notional "header" for the document (yay for overloading terms!)

app/controllers/manuals_controller.rb
```diff
@@ -6,33 +6,50 @@ class ManualsController < ApplicationController
...
 
   def show
     set_expiry(document_from_repository)
     document = Document.new(document_from_repository, manual)
+    presented_document = DocumentPresenter.new(document: document)
 
     render :show, locals: {
       presented_manual: presented_manual,
-      document: document
+      presented_document: presented_document,
```
h-lame: My comments on keeping `manual` instead of `previous_manual` apply here too: I don't think there's a compelling reason for the view to know this is a presented document.

app/views/layouts/application.html.erb
```diff
@@ -10,6 +10,7 @@
   <%= javascript_include_tag "application" %>
   <%= csrf_meta_tags %>
   <%= yield :extra_headers %>
+  <%= yield :extra_head_content %>
```
h-lame: I don't think we need this; `content_for` will collect all input given and yielding it outputs them all, so where we've got `content_for(:extra_head_content)` we can re-use `content_for(:extra_headers)` and it should be fine.  FWIW: I thing `extra_head_content` is a better name, so if you wanted to change `extra_headers` to `extra_head_content` I'd be into that.

app/views/manuals/index.html+new_navigation.erb
```diff
@@ -0,0 +1,15 @@
...
+  )
+%>
+<%= content_for :extra_head_content do %>
+  <%= ab_test.requested_variant.analytics_meta_tag.html_safe %>
+<% end %>
+<%= render 'govuk_component/beta_label' %>
+<%= render 'govuk_component/breadcrumbs',
+  breadcrumbs: presented_manual.taxonomy_breadcrumbs %>
+<%= render 'header', presented_manual: presented_manual %>
+<%= render 'manual', presented_manual: presented_manual %>
```
h-lame: Ah, I see now why we extracted `manual`, `manual_section`, and `manual_updates` it's so we can use them in these new a/b test commits.  Ignore my previous comment I suppose.  I reckon it might be good to flatten them down when/if we decide to keep the new navigation behaviour.

I've a feeling however that folk are going to want (particularly on the HMRC manuals which can be deeply nested as manual -> section -> section -> section) the manual breadcrumbs to be closer to the content (e.g. after the header).

app/presenters/document_presenter.rb
```diff
@@ -0,0 +1,45 @@
...
+      document_breadcrumbs.pop
+
+      # Re-add the manual entry, now as a regular breadcrumb
+      document_breadcrumbs << {
+        title: manual.title,
+        url: manual.url
+      }
+
+      # Add the section breadcrumb a `current`
+      document_breadcrumbs << {
```
h-lame: For "our" manuals (ones that go through manuals publisher) this works, as they can only be manual -> section, but for the hmrc manuals (ones that come from their publisher via hmrc-manauls-api) this won't work as it looks like it adds only the root (manual) and leaf (current document) nodes, but they can be arbitrarily nested within other documents (e.g. https://www.gov.uk/hmrc-internal-manuals/international-manual/intm162540 which has two intermediate parent documents between it and the manual).

I suspect this won't matter for the a/b test you are running as it's just education content, but it's probably safer to make sure we've got the right structure from the get go just in case.

app/views/manuals/_header.html.erb
carvil: 👍

app/views/manuals/_header.html.erb
carvil: I didn't know we could use that, TIL. I will make the change.

app/controllers/manuals_controller.rb
carvil: I understand your point and originally in an uncommitted version I did leave it as manual. My reasoning to change it to a `presented_manual` rather than `manual` is to make it easier to understand that there might be other methods we can use in the views (from `ManualPresenter`) rather than just what is defined in `Manual`.

For example, when we call `presented_manual.breadcrumbs`, it's clearer where that method definition is coming from.

But if you think it would be better to change it back to a `manual` I can do that.

app/controllers/manuals_controller.rb
carvil: Yup, I think my response above also applies here: https://github.com/alphagov/manuals-frontend/pull/198#discussion_r102446378

app/views/manuals/_content_for_manual.html.erb
carvil: That is a very good point! Since the content in that partial is to be included inside `<head>`, what do you think of renaming it to `_meta_content.html.erb`? Naming things is hard...

app/views/manuals/_content_for_manual.html.erb
carvil: Yup, that is a lot cleaner, I will make this change.

app/views/manuals/index.html.erb
carvil: As mentioned below, these changes are in preparation to reuse the partials in the new navigation to avoid repeating code that might get out of sync quickly during the beta of the new navigation. 

I refer to https://github.com/alphagov/manuals-frontend/pull/198#discussion_r102432199

app/views/layouts/application.html.erb
carvil: Ah, that's a good point. I will rename `extra_headers` to `extra_head_content` and remove the redundant one.

app/views/manuals/index.html+new_navigation.erb
carvil: Yeh, so the idea is when the new navigation is proven to be better than the existing one, we will be doing work to remove all this duplication. For now, we have to keep it as it is.

We are only rolling out the new navigation for Education content, which is the only Taxonomy we have developed so far. I don't think there will be changes to HMRC manuals, but I will check this with the team. Thanks for point it out.

app/presenters/document_presenter.rb
carvil: That's a good point. This won't impact this release, but I will take this feedback back to the team so we make sure we cater for those cases.

app/views/manuals/_content_for_manual.html.erb
h-lame: `meta_content` sounds good, clearer than `content_for_manual` anyway! :+1:

app/views/manuals/_content_for_manual.html.erb
```diff
@@ -1,6 +1,6 @@
 <% content_for :extra_headers do %>
-  <% if presented_manual.summary.present? %>
-    <meta name='description' content='<%= presented_manual.summary %>' />
+  <% if description.present? %>
+    <meta name='description' content='<%= description %>' />
```
h-lame: Now you've done this, you can use this `content_for_manual` partial in the `updates.html.erb` and `updates.html+new_navigation.erb` templates too.

app/views/manuals/updates.html.erb
h-lame: I can see why we wouldn't put the beta-label into the header partial (because we're changing the condition for showing the beta-label in the new_navigation versions).  I still think the partial is in the wrong place though on this page.

app/controllers/manuals_controller.rb
h-lame: Sure, I think the advantage for someone new reading the name outweighs my mild distaste for it.  It's a pretty minor thing so lets leave it as `presented_xxx`.

app/views/manuals/_content_for_manual.html.erb
carvil: @h-lame that makes sense. I've added a few more commits (including renaming that partial).

What do you think?

app/views/manuals/updates.html.erb
carvil: I blame GitHub for collapsing these comments and lot letting me see them in the view! I was going through the email and realised I missed a couple of your comments.

You are totally right, that beta label is out of place. Changing that now.

app/views/manuals/_content_for_manual.html.erb
h-lame: :+1:

app/views/manuals/updates.html.erb
h-lame: This is possibly my fault for doing it commit by commit.  Sorry!

PR #200, opened 2017-02-23
Column two thirds in manuals ab test

carvil:
This PR moves the manual pages into a `column-two-thirds` layout in order to make space for side navigation.

Trello: https://trello.com/c/V6PSacVC/336-changes-to-education-related-content-pages-as-part-of-the-new-navigation-in-manuals-frontend

The manual page:

screen shot 2017-02-23 at 10 35 37

The manual section page:

screen shot 2017-02-23 at 10 35 47

The manual updates' page:

screen shot 2017-02-23 at 10 35 57

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -16,14 +16,6 @@ main {
   display: block;
 }
 
-.outer-block {
-  @include outer-block;
-}
-
-.inner-block {
-  @include inner-block;
-}
-
 .manuals-frontend-body {
   padding-bottom: $gutter;
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -13,7 +13,7 @@ class ApplicationController < ActionController::Base
 private
 
   def slimmer_headers
-    slimmer_template "header_footer_only"
+    slimmer_template 'core_layout'
     set_slimmer_headers(remove_search: true)
   end
 end
--- a/app/views/manuals/index.html+new_navigation.erb
+++ b/app/views/manuals/index.html+new_navigation.erb
@@ -11,5 +11,13 @@
 <%= render 'govuk_component/beta_label' %>
 <%= render 'govuk_component/breadcrumbs',
   breadcrumbs: presented_manual.taxonomy_breadcrumbs %>
-<%= render 'header', presented_manual: presented_manual %>
-<%= render 'manual', presented_manual: presented_manual %>
+
+<div class='grid-row'>
+  <div class='column-two-thirds'>
+    <%= render 'header', presented_manual: presented_manual %>
+    <%= render 'manual', presented_manual: presented_manual %>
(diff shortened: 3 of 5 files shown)


h-lame:
This looks really weird without the side-navigation, can we hold off on doing this until the side navigation is present?  Squashing up the metadata `dl` in the header also looks quite strange - maybe the header should remain full-width but the content becomes 2/3rds ?  Might be worth getting some 2nd opinion from frontend / template consolidation / design folks here (obvs ignore me if you already have.)

carvil:
@h-lame this change only affects the B variant of the Education A/B test, which is only available on integration. We have other ongoing tasks to build the sidebar as a component and when that's ready we will plug it in.

This change does not affect any manual live at the moment and the change was reviewed by @alextea 

Happy to have a chat about all this if that's easier, please let me know.

fofr:
Is that breadcrumb usable?

It feels detached from the content now. Perhaps it's just the whitespace.

nickcolley:
We should only update master with complete changes, I don't think this change makes sense without implementing the side navigation?

h-lame:
One solution might be to move the "give feedback on this page" out of the header and into the side navigation.  That'd free up the whole of the 2/3rd width for the main contents of the header which would give everything in there some more room.

nickcolley:
@carvil let me know they're not deploying this variant until it's complete, so it makes sense to me 👍

carvil:
Apologies for the delay on the response, I was in a meeting.

As discussed IRL with @nickcolley these changes are part of the new navigation that we will launch in March. None of the `B` variants are live, they are only available on integration to allow for product and design review (it wouldn't be feasible to use only development as some team members don't have the development version of the site on their computers). The reason we aren't doing fewer PRs with more complete changes is because there are quite a lot of changes happening across several frontend applications. It wouldn't be practical from a team perspective to do it that way as we would be blocked very frequently.

This changes are are not the final changes for the new navigation. We will keep adding features to it and refining it as we test things. The entire set of changes will be product and design reviewed before the they are shown to actual users.

If you would like to checkout what we have been up to and what we are planning doing, it's on our trello board: https://trello.com/b/tYdZLGIe/navigation-team

We have done user research sessions in several areas of the new navigation, though not in everything (e.g. manuals). Both @fofr and @h-lame have good points in terms of usability and layout of the manual as-is. These will be tested out and @alextea is giving it some thought as well.

I'm sorry about the confusion, I should have explained a bit better what the intention with these changes is.

Does that make sense to you?

h-lame:
I hadn't quite realised the new navigation stuff was currently only on integration, so I'll be less wary when reviewing stuff in future.  Thanks for taking the time to explain everything though @carvil!  Apologies for causing any kerfuffle; I've approved this so you're good to merge.

carvil:
@h-lame no worries, I'm happy it's clearer now. Thanks for the feedback.

PR #206, opened 2017-03-02
Add taxonomy sidebar to new navigation pages

mobaig:
Add the taxonomy sidebar component from static and rework the tests to
assert for the presence of the sidebar where appropriate.

screen shot 2017-03-02 at 17 42 10

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,7 @@ gem 'gds-api-adapters', '40.1.0'
 gem 'airbrake', '4.0.0'
 gem 'logstasher', '0.6.2'
 
-gem 'govuk_navigation_helpers', '2.4.0'
+gem 'govuk_navigation_helpers', '~> 3.0'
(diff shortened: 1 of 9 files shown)


Gemfile
```diff
@@ -15,7 +15,7 @@ gem 'gds-api-adapters', '40.1.0'
 gem 'airbrake', '4.0.0'
 gem 'logstasher', '0.6.2'
 
-gem 'govuk_navigation_helpers', '2.4.0'
+gem 'govuk_navigation_helpers', '~> 3.0.0'
```
carvil: The commit message mentions version `3.0.1`, which is the latest. Should it be here?

mobaig: Actually I think this should be `3.0`. Updating.

mobaig:
@carvil updated. CI must have been glitchy.

PR #211, opened 2017-03-07
Only add A/B testing Vary header and meta tag to education pages

suzannehamilton:
Ensure that only pages in the navigation A/B test (those which are tagged to taxons) are cached separately and tracked in the A/B test analytics.

https://trello.com/c/Xehe42eM/464-only-add-a-b-meta-tag-to-pages-in-the-navigation-a-b-test

### DIFF
--- a/app/views/manuals/_content_for_head_tag.html.erb
+++ b/app/views/manuals/_content_for_head_tag.html.erb
@@ -2,6 +2,7 @@
   <% if description.present? %>
     <meta name='description' content='<%= description %>' />
   <% end %>
+  <%= ab_test.requested_variant.analytics_meta_tag.html_safe if add_ab_meta_tag %>
 <% end %>
 
 <% content_for :title, title %>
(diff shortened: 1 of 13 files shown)


app/views/manuals/_content_for_head_tag.html.erb
```diff
@@ -2,6 +2,7 @@
   <% if description.present? %>
     <meta name='description' content='<%= description %>' />
   <% end %>
+  <%= ab_test.requested_variant.analytics_meta_tag.html_safe if add_ab_meta_tag %>
```
alecgibson: Instead of using this `if` could we not use a variant? That would be more consistent with using variants elsewhere, and would remove the need for passing around the `add_ab_meta_tag` flag.

suzannehamilton: We discussed this in person, and we think we can't create a variant for whether a page is A/B tested or not, because it would conflict with the old/new variant.

I agree that the boolean parameter is ugly but it'll be removed at the end of the A/B test.

alecgibson: Yup, sorry got confused!

PR #214, opened 2017-03-08
Update navigation helpers gem to pick up the latest taxonomy sidebar

suzannehamilton:
The list of document types which appear in the sidebar has changed. This gem upgrade picks up the latest version.

https://trello.com/c/RhEocgyj/42-remove-notices-from-guidance-list

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -84,7 +84,7 @@ GEM
     govuk_frontend_toolkit (1.2.0)
       rails (>= 3.1.0)
       sass (>= 3.2.0)
-    govuk_navigation_helpers (3.1.0)
+    govuk_navigation_helpers (3.1.1)
       gds-api-adapters (~> 40.1)
     hashdiff (0.3.0)
     hike (1.2.3)


Gemfile
```diff
@@ -15,7 +15,7 @@ gem 'gds-api-adapters', '40.1.0'
 gem 'airbrake', '4.0.0'
 gem 'logstasher', '0.6.2'
 
-gem 'govuk_navigation_helpers', '~> 3.1'
+gem 'govuk_navigation_helpers', '~> 3.1.1'
```
mobaig: This isn't necessary - unless your intent was explicitly to ensure we only get bug fixes on `bundle update`?Despite recent versioning shenanigans on that gem, IMHO we should continue to try and conform to SemVer and have this line declare the dependency at the minor patch level.

suzannehamilton: Argh, I misremembered how `~>` works. I thought it always worked at a patch level, so `~> 3.1` and `~> 3.1.1` would both only allow bundler to install `3.1.x`, but the former actually allows `3.x`.

PR #217, opened 2017-03-14
Convert taxon breadcrumbs to back link on mobile

suzannehamilton:
Upgrade the govuk_navigation_helpers gem to pick up the latest breadcrumb config. This adds the taxon back link on mobile.

https://trello.com/c/Y7OirowB/472-mobile-version-of-breadcrumbs-should-only-show-previous-link


suzannehamilton:
@alecgibson I've fixed the tests and added the missing `collapse_on_mobile` flag, so it'd be good if you (or someone else) could re-review.

PR #225, opened 2017-04-04
Display assets correctly on development

rubenarakelyan:
This commit adds a development-specific configuration to ensure the frontend uses the assets subdomain for all assets rather than “www”.

Before:

screen shot 2017-04-04 at 12 16 47

After:

screen shot 2017-04-04 at 12 16 03

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -23,4 +23,7 @@ ManualsFrontend::Application.configure do
   # This option may cause significant delays in view rendering with a large
   # number of complex assets.
   config.assets.debug = true
+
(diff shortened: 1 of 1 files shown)


rubenarakelyan:
This is used by Rails methods like `stylesheet_link_tag` to link to the CSS/JS on the asset server rather than the same hostname that the app itself is accessed with. Since that doesn't use Plek, we need to set the Rails environment to the correct host.

rubenarakelyan:
I've changed this to be consistent with the production config, which doesn't use Plek but uses a different env var.

PR #227, opened 2017-04-06
Manual print page numbering

theleebriggs:
### Motivation
When printing `manual_sections` the browser generated list numbering on `legislative-lists` is doubling up with numbering / lettering put in the markup by publishers. This results in confusing list numbering in the print version.

Trello: https://trello.com/c/UibrijCb/6-2-immigration-manual-print-page-numbering

Example: https://www.gov.uk/guidance/immigration-rules/immigration-rules-appendix-fm-se-family-members-specified-evidence

### Description
This PR adds a `manuals-frontend` specific print style to hide the browser generated list numbers so that the publishers original numbering alone is visible in the print version.

### Before
before-list

### After
after-list

### DIFF
--- a/app/assets/stylesheets/_manuals-print.scss
+++ b/app/assets/stylesheets/_manuals-print.scss
@@ -64,3 +64,17 @@ a {
     font-weight: bold;
   }
 }
+
+// TODO:
+// legislative lists are being published with hardcoded item numbers
+// when printing this results in doubled up numbering.
+// These styles hide browser numbering on these lists.
+// Ideally this would be a global solution not just in manuals-frontend
+.govuk-govspeak .legislative-list {
+  padding-left: 0;
+
+  &,
+  ol {
+    list-style: none;
+  }
+}


app/assets/stylesheets/_manuals-print.scss
```diff
@@ -64,3 +64,17 @@ a {
...
   }
 }
+
+
+// TODO:
+// legislative lists are being published with hardcoded item numbers
+// when printing this results in doubled up numbering.
+// These styles hide browser numbering on these lists.
+// Ideally this would be a global solution not just in manuals-frontend
+.govuk-govspeak .legislative-list {
```
fofr: Should these print styles be in static? ie here: https://github.com/alphagov/static/blob/master/app/assets/stylesheets/govuk-component/_govspeak-print.scss

theleebriggs: That was something I thought about doing, I opted for limiting it to manuals rather than the govspeak component styles as we can't be sure on the way publishers are using them - so didn't want to globally strip the numbering, just limit it to cases we've seen where they are manually adding numbers / letters. Thoughts?

theleebriggs:
As per discussion these changes will be put into `static` so that they apply globally to the `legislative-list` component.

PR #229, opened 2017-05-24
Manual search box visual fixes

### LINKED ISSUE #183: Inconsistent search input rendering
Issue found while visiting the Scottish Government's Device Lab

Notice how when this was reimplemented a regression was introduced somewhere. (Potentially they were done at the same time, I'm not certain 😄 )

screen shot 2016-11-21 at 21 25 28
screen shot 2016-11-21 at 21 24 49

andysellick:
* Adjusted position of magnifying glass icon on submit button for high density screens
* Fixed height of search box on mobile to match height of submit button
* Adjusted height of button to even number for general height matching
* Fixes #183

**Before, desktop**
screen shot 2017-05-24 at 15 28 13

**After, desktop**
screen shot 2017-05-24 at 15 28 49

**Before, mobile**
screen shot 2017-05-24 at 15 29 26

**After, mobile**
screen shot 2017-05-24 at 15 30 02

Internet Explorer 9...

Before
screen shot 2017-05-25 at 08 41 48

After:
screen shot 2017-05-25 at 08 41 58

The only difference I've made is that the input and the button are now both a little taller. It looks exactly the same in every newer version of IE as well.

I can fix this by explicitly setting a px height for the input (allowing for padding), but that feels wrong.

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -132,12 +132,16 @@ main {
         width: calc(100% - 47px);
         // scss-lint:enable DuplicateProperty
         border: none;
-        padding: 6px;
+        padding: 7px;
+
+        @include media(mobile){
+          padding: 9px 7px;
+        }
       }
 
       button {
-        width: 33px;
-        height: 33px;
+        width: 34px;
+        height: 34px;
         border: 1px solid $manual-search-button-colour;
         border-width: 0 0 0 1px;
         border-left-color: $manual-search-button-border-colour;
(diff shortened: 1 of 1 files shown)


andysellick:
Before and even after all this it still isn't perfect in a lot of IE versions. We also seem to have a lot of search boxes that look the same but use different markup/CSS - would be good to get them all using the same component with better cross browser support at some point.

e.g.

https://www.gov.uk/
https://www.gov.uk/search
https://www.gov.uk/hmrc-internal-manuals/employment-income-manual

PR #230, opened 2017-06-05
Rails 5 upgrade

mcgoooo:
The only part of note in this is we had to make it explicit in Jasmine
to use the system version of phantomjs


RoryMacDonald:
Closing this, as have another PR that handles this

PR #234, opened 2017-07-17
Remove unneeded paragraph padding from collapsible sections

alecgibson:
Adding padding to the top of `p` elements in collapsible subsections
adds extra padding to govspeak components like callouts, warnings, step
numbers, etc. which makes them render incorrectly.

This margin should not be required elsewhere, because govspeak `p`
elements have bottom margin. The margin at the top is covered by
the margin introduced by the body wrapper.

### Before

callout-before
warning-before
numbers-before

### After

callout-after
warning-after
numbers-after

## Zendesk

https://govuk.zendesk.com/agent/tickets/2260672

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -303,10 +303,6 @@ main {
 
   .collapsible-subsections {
     @include media(tablet) {
-      p {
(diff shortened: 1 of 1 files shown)


fofr:
Does this have any effect on a paragraph that follows another paragraph?

PR #235, opened 2017-08-18
Fix caching problems and return 406 when necessary

thomasleese:
Slimmer is erroring under a certain set of inputs but catching the error – this leads to an invalid page being cached and served to users. By letting the error through we will serve a 500, the page will not be cached and users will see the correct content.

When a request is made for an unsupported format, eg text/javascript, the lack of a respond_to block meant we attempted to render it without a layout.

This led to a body being sent to slimmer without a #wrapper element,
causing the error:
`> undefined method 'to_html' for nil:NilClass`
in the BodyInserter processor

This matches: https://github.com/alphagov/finder-frontend/pull/271

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,8 +1,8 @@
 ManualsFrontend::Application.routes.draw do
-  with_options(constraints: { prefix: /(guidance|hmrc-internal-manuals)/ }) do |r|
-    r.get '/:prefix/:manual_id', to: 'manuals#index'
-    r.get '/:prefix/:manual_id/updates', to: 'manuals#updates'
-    r.get '/:prefix/:manual_id/:section_id', to: 'manuals#show'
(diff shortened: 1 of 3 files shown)


config/routes.rb
```diff
@@ -1,8 +1,10 @@
 ManualsFrontend::Application.routes.draw do
   with_options(constraints: { prefix: /(guidance|hmrc-internal-manuals)/ }) do |r|
-    r.get '/:prefix/:manual_id', to: 'manuals#index'
-    r.get '/:prefix/:manual_id/updates', to: 'manuals#updates'
-    r.get '/:prefix/:manual_id/:section_id', to: 'manuals#show'
+    scope format: true, constraints: { format: 'html' } do
```
kevindew: I'm not familiar with the `with_options` method but that looks like it's basically doing the same thing as scope and could be replaced and merged with the scope we've got rather than needing nested blocks?

PR #238, opened 2017-09-21
Replace airbrake with sentry

h-lame:
For: https://trello.com/c/FjUI2sc9/60-update-manuals-frontend


app/controllers/manuals_controller.rb
```diff
@@ -1,8 +1,4 @@
-require 'gds_api/helpers'
```
tijmenb: ❤️

PR #242, opened 2017-10-31
Upgrade to Rails 4.2.10

davidwinter:
Specs and Ruby linting is now passing and have verified this locally


h-lame:
Hi @davidwinter the easiest way to run the app against real data and apis is to use the development VM - all the details about that are here https://docs.publishing.service.gov.uk/manual/get-started.html

Gemfile
```diff
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-gem 'rails', '4.2.7.1'
+gem 'rails', '~> 4.2.10'
```
tijmenb: Can we pin this to the actual version like in other Rails app?

davidwinter:
We’re going to pull these lint changes out of this PR: https://github.com/alphagov/manuals-frontend/pull/242/files#diff-3c763ac78c7a85c70fd1f70e5382b1bf

We were scratching our heads on how the previous builds had passed the Jenkins builds, but seems like Jenkins only runs linting against the git diffs, which means there are probably offending files in a few repos. 

We’ll submit separate PR’s when we stumble across this in future, and pull these changes out of this PR.

@tijmenb ^

danielburnley:
The file with the lint changes was deleted on master in #243 - So this PR now only contains changes related to the rails update

PR #244, opened 2017-11-07
Initial Dockerfile

stephenharker:
Should build but may need some tweaks.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,20 @@
+FROM ruby:2.4.2
+RUN apt-get update -qq && apt-get upgrade -y
+
+RUN apt-get install -y build-essential nodejs && apt-get clean
+
+ENV GOVUK_APP_NAME manuals-frontend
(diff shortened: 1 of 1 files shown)


kevindew:
@RyanMacG Are you confident this works? it looks good to me

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+RUN mkdir $APP_HOME
+
+WORKDIR $APP_HOME
+ADD Gemfile* $APP_HOME/
+ADD .ruby-version $APP_HOME/
+RUN bundle install
+
+ADD . $APP_HOME
+
+CMD bash -c "bundle exec rails s -p $PORT -b '0.0.0.0'"
```
adrianclay: Prefixing with "rm -f tmp/pids/server.pid && " would be useful, as I managed to get a stray PID file while testing.

adrianclay:
I notice this will conflict with #248

PR #245, opened 2017-11-07
Add TaskList Header AB Testing

Davidslv:
This is only to render the tasklist header on the "/guidance/the-highway-code"

Trello: https://trello.com/c/XGpc4qlo

### DIFF
--- a/app/views/manuals/index.html.erb
+++ b/app/views/manuals/index.html.erb
@@ -5,7 +5,15 @@
     title: presented_manual.full_title,
   )
 %>
-<% if presented_manual.beta? %>
+
+<% if show_tasklist_header? %>
+  <%= render "govuk_component/task_list_header", {
+    title: "Learn to drive a car: step by step",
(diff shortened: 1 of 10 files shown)


app/views/manuals/index.html.erb
```diff
@@ -5,7 +5,15 @@
     title: presented_manual.full_title,
   )
 %>
-<% if presented_manual.beta? %>
+
+<% if show_tasklist_header? %>
+  <%= render "govuk_component/task_list_header", {
+    title: "Learn to drive a car: step by step",
+    path: "/learn-to-drive-a-car"
```
deborahchua: We need to add the skip link details:
 `skip_link: "#check-you're-allowed-to-drive",
  skip_link_text: "Check you're allowed to drive"`

Davidslv: This page is a full width page, so there isn't a sidebar tasklist, meaning we don't "skip_link" this one.

PR #282, opened 2018-01-05
Bump components gem to 3.2.0

andysellick:
Mainly for the new print styles for task lists.
  
Trello card: https://trello.com/c/9CD3dY7r/409-add-task-list-print-styles-to-apps


andysellick:
@sihugh oh yeah that's a good point. Shall we bother with this PR at all then?

sihugh:
Yeah, it saves a dependabot PR in about 5 minutes :-)

PR #292, opened 2018-01-16
Upgrade to govuk_app_config v 1.2.1

h-lame:
We follow the guide in the changelog for govuk_app_config for going to
1.0.  We also remove the following stanza from config/production.rb:

    $real_stdout = $stdout.clone
    $stdout.reopen($stderr)

It's not mentioned in the govuk_app_config readme, but those lines are in
govuk_app_config now, so I suspect we don't need them in ours.


h-lame:
@sihugh - good catch, fixed in a rebase.

PR #306, opened 2018-02-06
Return a redirect when the content-store returns one

adrianclay:
When a section is withdrawn from a manual it gets saved as a redirect in content-store.

There exists a race condition where the router retrieves the content as being a manual, then before the manuals-frontend requests the document, publishing-api updates the type to redirect.  This race condition happens relatively often in the E2E tests.

This change will redirect the request properly instead of raising an exception inside the Manual model.


h-lame:
I think we've shied away from having the frontend apps know about redirects and gones on the assumption that the router should deal with it.  Obviously it doesn't always so I think this is pragmatic in the face of errors and a race-condition.  I wonder if other content-store-based frontend apps have the same problem? Have we fixed it there, and if not, should we?  Is this something we can fix in gds-api-adapters or some other gem we pull into all our frontend apps?

adrianclay:
Doh, should have linked out.

Main discussion of this is here: https://github.com/alphagov/gds-api-adapters/pull/805

Similar fix for government-frontend here: https://github.com/alphagov/government-frontend/pull/727

Other frontends could be susceptible to similar problems around this.  These are the two frontends which have been blowing up on E2E tests.  Admittedly we don't have tests which would blow up like this for [alphagov/frontend](https://www.github.com/alphagov/frontend)

app/repositories/document_repository.rb
```diff
@@ -31,8 +31,9 @@ def extract_parent_base_path_from_document(document)
   end
 
   def fetch_parent_for_document(document)
-    fetch_content_item(
-      extract_parent_base_path_from_document(document)
-    )
+    parent = extract_parent_base_path_from_document(document)
+    if parent
+      fetch_content_item(parent)
```
kevindew: might be a little nicer to have this as `fetch_content_item(parent) if parent`

kevindew:
@h-lame yeah it is a problem shared by other frontend apps and I'd agree it'd be good to handle it outside frontend apps - where all of them all already share various bits and pieces for dealing with content store responses, there must already be a lot of duplication.

I'm not sure of a particularly trivial way to resolve it though. We'd want to be in a situation where an app can give input of a path for a content item and a query string and then for that to either return a content item or something that could be used to make a response (which is what we pretty much have now). I guess what we'd ideally want is a way for a helper method to be able to return that response if it's an unexpected one whereas continue with it's execution if not. Best way I can think of for that would be some kind of exception with a rescue handler that comes from GdsApi or similar. Have you got any ideas? (though I don't expect it'll be developed imminently more something for tech debt board)

h-lame:
@kevindew I think you could wrap the solution from government-frontend up in a gem.  Have the content-store api adapter interrogate the response and raise a Redirect or Gone exception (with the contents of the response) and then have `<something>`* register a `rescue_from` in the ApplicationController** for those Redirect and Gone exceptions and then do the right thing.

* I'm not sure this belongs in gds-api-adapters as it's not got any other rails controller stuff in it
** Maybe this is opt-in, like you have to include a `<something>::ContentStoreExceptionHandler` in your ApplicationController, or maybe it's something that gloms onto it automatically when you load the gem.  Unsure TBH.

PR #307, opened 2018-02-07
Update step by step navigation

sihugh:
This PR:

- removes the AB test for task lists
- renames task lists to step by step nav
- uses the updated step by step nav components
- uses the data from navigation_helpers to determine when to show the step nav header

https://trello.com/c/3pQPJDmj/440-allow-collections-to-render-task-lists-from-the-publishing-api


Gemfile
```diff
@@ -10,8 +10,8 @@ gem 'uglifier', '>= 1.3.0'
 gem 'plek', '2.1.1'
 gem 'gds-api-adapters', '~> 51.2'
 
-gem 'govuk_publishing_components', '~> 4.1.1'
-gem 'govuk_navigation_helpers', '8.2.2'
+gem 'govuk_publishing_components', '~> 5.0'
```
andysellick: I'm not certain of the syntax here, presumably 5.0 is okay, it doesn't need to be 5.0.0? What if 5.0.1 gets released, will it still use 5.0.0?

sihugh: It'll be locked to the version in the Gemfile.lock unless someone (like dependabot) does a `bundle update`.  That syntax means that `bundle update` will try and update to the latest v5.*, which is probably fine given that we try to not break stuff with our minor releases.

PR #331, opened 2018-03-06
Render step nav content from content item

sihugh:
Step by step navigation content is now held in content items.  The helpers have
been updated appropriately, and are now held in govuk_publishing_components as
they are so tightly coupled to the components themselves.

https://trello.com/c/aHTclade/443-manuals-frontend-renders-step-by-step-navigation-from-content-item

### DIFF
--- a/app/views/manuals/index.html.erb
+++ b/app/views/manuals/index.html.erb
@@ -6,14 +6,11 @@
   )
 %>
 
-<% if show_step_nav? %>
-  <%= render "govuk_publishing_components/components/step_by_step_nav_header", {
-    title: current_step_nav.title,
-    path: current_step_nav.base_path
-  } %>
+<% if step_nav_helper.show_header? %>
+  <%= render "govuk_publishing_components/components/step_by_step_nav_header", step_nav_helper.header %>
 <% end %>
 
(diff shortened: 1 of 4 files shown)


app/views/manuals/index.html.erb
```diff
@@ -6,14 +6,11 @@
   )
 %>
 
-<% if show_step_nav? %>
-  <%= render "govuk_publishing_components/components/step_by_step_nav_header", {
-    title: current_step_nav.title,
-    path: current_step_nav.base_path
-  } %>
+<% if step_nav_helper.show_header? %>
```
Davidslv: This will not work if step_nav_helper is nil

sihugh: I don't think it should be nil, as it'll instantiate a new one when called (see the application controller).  I've got tests for when links are/aren't present, so I think this should be fine?

sihugh: I don't think it should be nil, as it'll instantiate a new one when called (see the application controller).  I've got tests for when links are/aren't present, so I think this should be fine?

I'll rebase too, thanks

Davidslv: Got it! Thanks :)

PR #392, opened 2018-05-30
Add schema and SEO things to manuals

tijmenb:
This adds the Machine readable metadata component, which adds Schema.org data, a canonical tag and Twitter/Facebook meta tags.

https://trello.com/c/gKHoUGuA

### DIFF
--- a/app/views/manuals/_content_for_head_tag.html.erb
+++ b/app/views/manuals/_content_for_head_tag.html.erb
@@ -2,6 +2,15 @@
   <% if description.present? %>
     <meta name='description' content='<%= description %>' />
   <% end %>
+
+  <% if @content_store_section %>
+    <%= render 'govuk_publishing_components/components/machine_readable_metadata',
+      schema: :article,
+      content_item: @content_store_section,
+      title: title,
+      description: description
+    %>
+  <% end %>
 <% end %>
 
 <% content_for :title, title %>
--- a/app/controllers/manuals_controller.rb
+++ b/app/controllers/manuals_controller.rb
@@ -73,6 +73,14 @@ private
     end
   end
 
+  def content_store_section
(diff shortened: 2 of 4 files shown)


app/views/manuals/_content_for_head_tag.html.erb
```diff
@@ -2,6 +2,15 @@
   <% if description.present? %>
     <meta name='description' content='<%= description %>' />
   <% end %>
+
+  <%= render 'govuk_publishing_components/components/machine_readable_metadata',
+    schema: :article,
+    content_item: @content_store_manual,
```
sihugh: What's the reasoning behind passing the manual content item rather than the section?

tijmenb: Ow, I missed that. I'll take a look.

tijmenb:
Closing until Wednesday, when I'll take another look.

app/views/manuals/_content_for_head_tag.html.erb
tijmenb: @sihugh updated to be the section instead of manual. Thanks!

app/views/manuals/_content_for_head_tag.html.erb
```diff
@@ -2,6 +2,15 @@
   <% if description.present? %>
     <meta name='description' content='<%= description %>' />
   <% end %>
+
+  <% if @content_store_section %>
```
Davidslv: Would be better not to use instance variables in partials, instead passed them through as local variables of this partial.

- http://www.carlosramireziii.com/stop-using-instance-variables-in-partials.html
- https://forum.upcase.com/t/best-practice-dont-use-instance-variables-in-partials/2004

PR #404, opened 2018-06-15
Use govspeak component from gem

tijmenb:
https://trello.com/c/ZMVwIyj5/28-move-the-govspeak-component-to-the-gem

### DIFF
--- a/app/models/section.rb
+++ b/app/models/section.rb
@@ -8,7 +8,7 @@ class Section
   end
 
   def body
-    section['body']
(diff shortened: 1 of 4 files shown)


rubenarakelyan:
Does all content need `html_safe`?

tijmenb:
Yeah, it's govspeak coming from the content store. I did notice that in most cases the helper does `.html_safe` already though, so I've reworked the commit to be consistent about that.

PR #531, opened 2018-12-04
Override link colour in metadata component

andysellick:
- assumes a white background but the component is used on a blue background so links are invisible
- this is a short term fix while we get the component updated to have an option to allow for this scenario

Fixes this:

screen shot 2018-12-04 at 10 47 42

Example demo URL (fixed): https://govuk-manuals-frontend-pr-531.herokuapp.com/guidance/prevent-groundwater-pollution-from-underground-fuel-storage-tanks

Example live URL (broken): https://www.gov.uk/guidance/prevent-groundwater-pollution-from-underground-fuel-storage-tanks

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -27,6 +27,18 @@ header .gem-c-metadata {
   header {
     background: $govuk-blue;
 
+    // FIXME this is to fix link colours in the metadata component on a
+    // blue background. This can be removed once the component has been
+    // updated to accommodate such a situation
+    .gem-c-metadata a,
+    .gem-c-metadata a:link,
+    .gem-c-metadata a:hover,
+    .gem-c-metadata a:active,
+    .gem-c-metadata a:visited {
+      color: govuk-colour('white');
+    }
+
+
     &.hmrc {
       background: $hm-revenue-customs;
     }


app/assets/stylesheets/application.scss
```diff
@@ -27,6 +27,18 @@ header .gem-c-metadata {
...
 
+    // FIXME this is to fix link colours in the metadata component on a
+    // blue background. This can be removed once the component has been
+    // updated to accommodate such a situation
+    .gem-c-metadata a,
+    .gem-c-metadata a:link,
+    .gem-c-metadata a:hover,
+    .gem-c-metadata a:active,
+    .gem-c-metadata a:visited {
+      color: $white;
```
DilwoarH: should you use the govuk-frontend colours 

`govuk-colour('white')`

app/assets/stylesheets/application.scss
```diff
@@ -27,6 +27,18 @@ header .gem-c-metadata {
   header {
     background: $govuk-blue;
 
+    // FIXME this is to fix link colours in the metadata component on a
+    // blue background. This can be removed once the component has been
+    // updated to accommodate such a situation
+    .gem-c-metadata a,
+    .gem-c-metadata a:link,
+    .gem-c-metadata a:hover,
```
elliotcm: Do we need to maintain a distinct hover state colour, or is it OK to have them all be white like this?

app/assets/stylesheets/application.scss
andysellick: Not sure if that'll work in this app, but I'll give it a go.

app/assets/stylesheets/application.scss
andysellick: I did wonder about that, but I don't know what the hover colour should be - this has to work on a dark blue background and also a teal-ish background. Changing link colour on hover isn't an accessibility requirement (I think) so I think for now it's easier if we just leave them white.

app/assets/stylesheets/application.scss
andysellick: It worked, thanks.

PR #642, opened 2019-05-16
Fix accordion accessibility

kr8n3r:
Some issues have been pointed out with the implementation of the accordion functionality on the following page template:
https://www.gov.uk/guidance/how-to-publish-on-gov-uk/accessible-pdfs

It's an improvement on the previous implementation but not as accessible (to screen readers) as the implementation in the service manual e.g https://www.gov.uk/service-manual/design

Ideally we'd use the https://govuk-publishing-components.herokuapp.com/component-guide/accordion component but as this content is generated with govspeak, that isn't possible.

This PR adds fixes present in the above linked component

- Updates markup to use `button` instead of `a`
- Adds aria-expanded attribute to the button to announce state change to screenreaders
- Adds a handleClick function that contains section toggle
and button attribute toggle to announce state change to screenreaders
- Replaces open/close all anchors with more semantic html (buttons)
- Replaces section header anchors with more semantic html (buttons)
- Adds functionality that toggles section button aria attribute on open/close all - to announce state change to screenreaders
- Adds functionality that toggles open/close button's disabled attribute

### DIFF
--- a/app/assets/javascripts/modules/collapsible.js
+++ b/app/assets/javascripts/modules/collapsible.js
@@ -5,11 +5,21 @@
   function Collapsible(section){
     this.$section = section;
     this.$clickTarget = this.$section.find('.js-subsection-title');
-    this.$clickTarget.on('click', this.toggle.bind(this));
+    this.$clickTarget.on('click', this.handleClick.bind(this));
+    this.$button = this.$clickTarget[0].querySelector('button');
     this.addToggle();
 
     this.$section.on('focus', '.js-subsection-body a', this.showSectionWhenLinkFocused.bind(this));
   }
+
+  Collapsible.prototype.handleClick = function handleClick(){
+    this.toggle()
+    this.toggleAriaExpanded()
+  };
+
+  Collapsible.prototype.toggleAriaExpanded = function toggleAriaExpanded() {
+    this.$button.getAttribute('aria-expanded') === 'false' ?  this.$button.setAttribute('aria-expanded', 'true') : this.$button.setAttribute('aria-expanded', 'false');
+  };
 
   Collapsible.prototype.showSectionWhenLinkFocused = function() {
     if (this.$section.is('.closed')) {
@@ -18,21 +28,22 @@
   };
 
   Collapsible.prototype.addToggle = function addToggle(){
-    var $toggleHTML = $("<span class='js-toggle'></span>");
+    var $toggleHTML = $("<span class='js-toggle' aria-hidden='true'></span>");
     this.$clickTarget.append($toggleHTML);
   };
 
-  Collapsible.prototype.toggle = function toggle(event){
+  Collapsible.prototype.toggle = function toggle(){
     this.$section.toggleClass('closed');
-    event.preventDefault();
   };
 
   Collapsible.prototype.close = function close(){
     this.$section.addClass('closed');
+    this.$button.setAttribute('aria-expanded', 'false');
   };
 
   Collapsible.prototype.open = function open(){
     this.$section.removeClass('closed');
+    this.$button.setAttribute('aria-expanded', 'true');
   };
 
   Collapsible.prototype.isClosed = function(){
--- a/app/assets/javascripts/modules/collapsible_collection.js
+++ b/app/assets/javascripts/modules/collapsible_collection.js
@@ -18,8 +18,8 @@
 
     if(this.$sections.length > 0) {
       this.$sections.each(this.initCollapsible.bind(this));
-      this.$openAll = $("<a href='#' aria-hidden=true>Open all</a>");
-      this.$closeAll = $("<a href='#' aria-hidden=true>Close all</a>");
+      this.$openAll = $("<button>Open all</button>");
+      this.$closeAll = $("<button>Close all</button>");
       this.addControls();
 
       this.closeAll();
@@ -65,7 +65,7 @@
     var subsectionHeaders = this.$container.find(this.collapseSelector);
     subsectionHeaders.each(function() {
       var $header = $(this);
-      $header.addClass('js-subsection-title').wrapInner('<a role="button" href="#' + $header.attr('id') + '"></a>');
+      $header.addClass('js-subsection-title').wrapInner('<button type="button" aria-expanded="false"></button>');
     });
 
     subsectionHeaders.each(function(index, el){
@@ -96,30 +96,22 @@
     return selector;
   };
 
-  CollapsibleCollection.prototype.closeAll = function closeAll(event){
+  CollapsibleCollection.prototype.closeAll = function closeAll(){
     for (var section in this.collapsibles) {
       this.collapsibles[section].close();
     }
 
     this.disableControl(this.$closeAll);
     this.enableControl(this.$openAll);
-
-    if (typeof event != 'undefined'){
-      event.preventDefault();
-    }
   };
 
-  CollapsibleCollection.prototype.openAll = function openAll(event){
+  CollapsibleCollection.prototype.openAll = function openAll(){
     for (var section in this.collapsibles) {
       this.collapsibles[section].open();
     }
 
     this.disableControl(this.$openAll);
     this.enableControl(this.$closeAll);
-
-    if (typeof event != 'undefined'){
-      event.preventDefault();
-    }
   };
 
   CollapsibleCollection.prototype.addControls = function addControls(){
@@ -161,11 +153,11 @@
   };
 
   CollapsibleCollection.prototype.disableControl = function disableControl(control){
-    control.addClass('disabled');
+    control.addClass('disabled').attr('disabled', 'disabled');
   };
 
   CollapsibleCollection.prototype.enableControl = function enableControl(control){
-    control.removeClass('disabled');
+    control.removeClass('disabled').removeAttr('disabled');
   };
 
   CollapsibleCollection.prototype.openCollapsibleForAnchor = function openCollapsibleForAnchor(anchor){
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -13,6 +13,16 @@
 $manual-search-button-colour: #0b0c0c;
 $manual-search-button-border-colour: #222;
 
+%common-button-styles  {
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
+  font-size: 100%;
+  border-width: 0;
+  color: $govuk-blue;
+  background: none;
+  cursor: pointer;
+}
+
 main {
   display: block;
 }
@@ -278,8 +288,15 @@ main {
     margin: $gutter 0;
     position: relative;
 
-    a {
-      text-decoration: none;
+    button {
+      @extend %common-button-styles;
+      font-weight: 700;
+      margin-top: 0;
+      margin-bottom: 0;
+      margin-left: 0;
+      padding-bottom: 0;
+      padding-left: 0;
+      text-align: left;
     }
   }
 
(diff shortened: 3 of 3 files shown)


app/assets/javascripts/modules/collapsible.js
```diff
@@ -5,26 +5,35 @@
...
-    this.$clickTarget.on('click', this.toggle.bind(this));
+    this.$clickTarget.on('click', this.handleClick.bind(this));
     this.addToggle();
 
     this.$section.on('focus', '.js-subsection-body a', this.showSectionWhenLinkFocused.bind(this));
   }
 
+  Collapsible.prototype.handleClick = function handleClick(){
+    this.toggle()
+    this.toggleAttribute()
```
ChrisBAshton: Could this be renamed to better convey what it's doing?

```js
this.toggleAriaExpanded()
```

app/assets/javascripts/modules/collapsible.js
```diff
@@ -5,26 +5,35 @@
...
+  };
+
+  Collapsible.prototype.toggleAttribute = function() {
+    var $button = $(this.$clickTarget[0].firstChild)
```
ChrisBAshton: If we pass the `event` from `handleClick`, we can avoid this slightly brittle looking code:

```js
this.toggleAttribute(e.target)

// ...

Collapsible.prototype.toggleAttribute = function(button) {
```

app/assets/javascripts/modules/collapsible.js
```diff
@@ -5,26 +5,35 @@
...
+
   Collapsible.prototype.showSectionWhenLinkFocused = function() {
     if (this.$section.is('.closed')) {
       this.$section.toggleClass('closed');
     }
   };
 
   Collapsible.prototype.addToggle = function addToggle(){
-    var $toggleHTML = $("<span class='js-toggle'></span>");
+    var $toggleHTML = $("<span class='js-toggle' aria-hidden='true'></span>");
```
ChrisBAshton: Could you explain why you've added `aria-hidden`? (And do so in the commit - I couldn't find an explanation...?)

app/assets/javascripts/modules/collapsible_collection.js
```diff
@@ -18,8 +18,8 @@
 
     if(this.$sections.length > 0) {
       this.$sections.each(this.initCollapsible.bind(this));
-      this.$openAll = $("<a href='#' aria-hidden=true>Open all</a>");
-      this.$closeAll = $("<a href='#' aria-hidden=true>Close all</a>");
+      this.$openAll = $("<button type='button'>Open all</button>");
+      this.$closeAll = $("<button type='button'>Close all</button>");
```
ChrisBAshton: 🎉 Thanks for this!

(First time I've seen `type="button"` - think I usually just do `<button>`, and provided it's in a form, it doesn't need `preventDefault`. I guess adding `type="button"` means you've been able to remove the `preventDefault` here: https://github.com/alphagov/manuals-frontend/pull/642/files#diff-8f262d6c492018c739c12c25a4a34efeL27 ?)

app/assets/javascripts/modules/collapsible_collection.js
```diff
@@ -98,28 +98,26 @@
 
   CollapsibleCollection.prototype.closeAll = function closeAll(event){
     for (var section in this.collapsibles) {
+      var $button = $(this.collapsibles[section].$clickTarget[0].firstChild)
```
ChrisBAshton: Does this work? Following the logic from line 25, where `this.closeAll()` is called with no 'event' in the object constructor (rather than in response to a click) I'd expect a `Cannot get firstChild of undefined` error here 🤔

app/assets/javascripts/modules/collapsible_collection.js
```diff
@@ -161,11 +159,11 @@
   };
 
   CollapsibleCollection.prototype.disableControl = function disableControl(control){
-    control.addClass('disabled');
+    control.addClass('disabled').attr('disabled','disabled');
```
ChrisBAshton:
```suggestion
    control.addClass('disabled').attr('disabled', 'disabled');
```

app/assets/stylesheets/application.scss
```diff
@@ -312,10 +326,17 @@ main {
...
+        button {
           @include core-16;
           padding: 0  5px 0 5px;
           display: inline-block;
+          -webkit-font-smoothing: antialiased;
+          -moz-osx-font-smoothing: grayscale;
+          font-size: 100%;
+          border-width: 0;
+          color: $govuk-blue;
+          background: none;
```
ChrisBAshton: Fair bit of repetition between this block and the [`button` block on line 281](https://github.com/alphagov/manuals-frontend/pull/642/files#diff-f859e2848b07324f808ce1a5ccd9fcbaR281) - is there any way those can be combined?

app/assets/stylesheets/application.scss
```diff
@@ -312,10 +326,17 @@ main {
...
+          cursor: pointer;
```
ChrisBAshton: Is `cursor: pointer` really necessary on a button? I don't know of any browser that wouldn't do this by default.

app/assets/stylesheets/application.scss
```diff
@@ -312,10 +326,17 @@ main {
           display: inline-block;
         }
 
-        a {
+        button {
           @include core-16;
           padding: 0  5px 0 5px;
           display: inline-block;
+          -webkit-font-smoothing: antialiased;
+          -moz-osx-font-smoothing: grayscale;
```
ChrisBAshton: I'm not sure where GOV.UK stands on vendor prefixes - will leave that to someone else to confirm (I'd like to know too!) 😄

app/assets/stylesheets/application.scss
kr8n3r: not necessary, as you could argue that pointer indicates a navigation action. by adding pointer i'm being consistent with https://design-system.service.gov.uk/components/accordion/ and https://govuk-publishing-components.herokuapp.com/component-guide/accordion

app/assets/stylesheets/application.scss
kr8n3r: going by the new typography styles in [govuk-frontend ](https://github.com/alphagov/govuk-frontend/blob/master/src/helpers/_typography.scss#L15)

I know this app is still using the (now deprecated) govuk_frontend_toolkit. I can remove the added prefixes, no problem

app/assets/stylesheets/application.scss
kr8n3r: thanks, will add shared styles in a placeholder

app/assets/javascripts/modules/collapsible.js
kr8n3r: done

app/assets/javascripts/modules/collapsible.js
kr8n3r: done in the commit

app/assets/javascripts/modules/collapsible.js
kr8n3r: it's worth remembering users can also click the toggle span, in which case the target is different but we still need to only toggle the button's attribute

kr8n3r: it's worth remembering users can also click the toggle span, in which case the target is different but we still need to only toggle the button's attribute

app/assets/javascripts/modules/collapsible_collection.js
kr8n3r: yeah, it works as the entire collapse block is passed through. i've removed the leftover event as i've already removed the preventDefault

app/assets/javascripts/modules/collapsible_collection.js
kr8n3r: you're correct that type=button isn't strictly necessary as it's unlikely this block will ever be in a form. Will remove as i've copied it over from https://govuk-publishing-components.herokuapp.com/component-guide/accordion

app/assets/javascripts/modules/collapsible.js
kr8n3r: replaced with a less brittle querySelector has the parent H2 is already available as the $clickTarget

app/assets/javascripts/modules/collapsible_collection.js
kr8n3r: type=button removed

app/assets/javascripts/modules/collapsible_collection.js
kr8n3r: also replaced with querySelector as we have the parent h2 available as the $clickTarget

app/assets/javascripts/modules/collapsible_collection.js
```diff
@@ -96,30 +96,30 @@
...
   };
 
-  CollapsibleCollection.prototype.closeAll = function closeAll(event){
+  CollapsibleCollection.prototype.closeAll = function closeAll(){
     for (var section in this.collapsibles) {
+      // as $clickTarget is the whole h2 and is available, we just need to get the button
+      var $button = $(this.collapsibles[section].$clickTarget[0].querySelector('button'))
+
       this.collapsibles[section].close();
+      $button.attr('aria-expanded', 'false');
```
ChrisBAshton: Can we remove the changes in the `openAll` and `closeAll` functions and instead encapsulate the behaviour in `collapsible.js`?

i.e. `this.collapsibles[section].close()` should, inside its function definition, ensure that the button `aria-expanded` is `false`.

kr8n3r: moved into collapsible and trimmed down

app/assets/javascripts/modules/collapsible.js
```diff
@@ -5,34 +5,46 @@
...
   };
 
   Collapsible.prototype.close = function close(){
     this.$section.addClass('closed');
+    this.$clickTarget[0].querySelector('button').setAttribute('aria-expanded', 'false');
   };
 
   Collapsible.prototype.open = function open(){
     this.$section.removeClass('closed');
+    this.$clickTarget[0].querySelector('button').setAttribute('aria-expanded', 'true');
```
ChrisBAshton: This can be cached in the constructor now I think? Would make the rest of the code much easier to read 👍 

```js
this.$button = this.$clickTarget[0].querySelector('button');
```

kr8n3r: done

kr8n3r:
closing in favour of #650 to get tests passing

app/assets/javascripts/modules/collapsible.js
```diff
@@ -5,34 +5,45 @@
...
+  Collapsible.prototype.handleClick = function handleClick(){
+    this.toggle()
+    this.toggleAriaExpanded()
+  };
+
+  Collapsible.prototype.toggleAriaExpanded = function toggleAriaExpanded() {
+    this.$button.getAttribute('aria-expanded') === 'false' ?  this.$button.setAttribute('aria-expanded', 'true') : this.$button.setAttribute('aria-expanded', 'false');
+  };
+
   Collapsible.prototype.showSectionWhenLinkFocused = function() {
```
tombye:
```suggestion
  Collapsible.prototype.showSectionWhenLinkFocused = function showSectionWhenLinkFocused() {
```

app/assets/javascripts/modules/collapsible.js
```diff
@@ -5,34 +5,45 @@
...
     this.$section.addClass('closed');
+    this.$button.setAttribute('aria-expanded', 'false');
   };
 
   Collapsible.prototype.open = function open(){
     this.$section.removeClass('closed');
+    this.$button.setAttribute('aria-expanded', 'true');
   };
 
   Collapsible.prototype.isClosed = function(){
```
tombye:
```suggestion
  Collapsible.prototype.isClosed = function isClosed(){
```

kr8n3r:
@tombye thanks, I'll add them to #642

PR #708, opened 2019-08-19
Update to publishing components v20

kr8n3r:
## What
The upgrade to version 20 of the component gem will bring in the changes made to version 3 of GOV.UK Design System.

The legacy template support has been turned on.

## Why
The improvements include a new focus state, which will help this app to reach the WCAG 2.1 standard. It also has a tonne of other improvements that cascade from Design System, such as the new smaller font.

## Visual changes
There are minor spacing changes that stem from typography/heading changes.
We could adjust these back with custom CSS, but I'm not convinced it's worth it

### Header title with caption
A bit less of spacing above and below the caption.
Before:
Screen Shot 2019-08-20 at 11 04 58

After: 
Screen Shot 2019-08-20 at 11 11 59

### Summary heading
Medium heading from GOV.UK Frontend has a bigger bottom margin.

Before:
Screen Shot 2019-08-20 at 11 13 57

After:
Screen Shot 2019-08-20 at 11 13 37

URLs to check:

https://govuk-manuals-frontend-pr-708.herokuapp.com/guidance/style-guide/

https://govuk-manuals-frontend-pr-708.herokuapp.com/hmrc-internal-manuals/pensions-tax-manual

https://govuk-manuals-frontend-pr-708.herokuapp.com/guidance/how-to-publish-on-gov-uk/accessible-pdfs


app/assets/stylesheets/application.scss
```diff
@@ -1,95 +1,92 @@
...
-@import "typography";
-@import "measurements";
-@import "colours";
-@import "shims";
-@import "css3";
-@import "grid_layout";
-
-@import "reset";
+$govuk-compatibility-govuktemplate: true;
+$govuk-typography-use-rem: false;
```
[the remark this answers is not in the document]
kr8n3r: 👍

app/assets/stylesheets/_manuals-print.scss
```diff
@@ -60,7 +60,11 @@ a {
   margin-bottom: 30px;
 
   h2 {
-    @include core-16;
+    font-family: "GDS Transport", Arial, sans-serif;
```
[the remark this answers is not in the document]
kr8n3r: was going on all other print style. I've changed it to 'nta' as template uses old font anyway

kr8n3r:
PR review app needs a kickstart

app/views/layouts/application.html.erb
```diff
@@ -3,8 +3,6 @@
 <head>
   <title><%= yield :title %> - GOV.UK</title>
   <!--[if gt IE 8]><!--><%= stylesheet_link_tag "application" %><!--<![endif]-->
-  <!--[if IE 6]><%= stylesheet_link_tag "application-ie6" %><script>var ieVersion = 6;</script><![endif]-->
-  <!--[if IE 7]><%= stylesheet_link_tag "application-ie7" %><script>var ieVersion = 7;</script><![endif]-->
```
andysellick: These scripts also need to be removed from https://github.com/alphagov/manuals-frontend/blob/master/config/initializers/assets.rb

andysellick: I know we don't support these browsers anymore but have you checked what effect removing these stylesheets has? I've faced a similar problem with the gov.uk homepage and have concluded that I'm not quite ready to break it for what turns out to be something like 40,000 users just yet.

app/assets/stylesheets/_manuals-print.scss
```diff
@@ -60,7 +60,11 @@ a {
   margin-bottom: 30px;
 
   h2 {
-    @include core-16;
```
andysellick: Sorry, confused. Why is this not being replaced with a `govuk-font` mixin?

app/assets/stylesheets/application.scss
```diff
@@ -1,95 +1,91 @@
...
+    background: $govuk-brand-colour;
 
     &.hmrc {
-      background: $hm-revenue-customs;
+      background: govuk-organisation-colour("hm-revenue-customs", $websafe: false);
     }
 
-    padding: $gutter;
-    color: $white;
+    padding: govuk-spacing(6) govuk-spacing(3);
```
andysellick: This is a different padding from the original. Is that deliberate?

Edit: oh, I think I understand, this is to do with the change below to the media queries?

app/assets/stylesheets/application.scss
```diff
@@ -1,95 +1,91 @@
...
 
-    @include media(tablet) {
       .primary,
       .secondary {
         float: left;
       }
 
       .primary {
-        width: $two-thirds;
+        width: govuk-grid-width("two-thirds");
```
andysellick: I appreciate this would be more work, but at some point we should clear out these grid styles and use the govuk-frontend grid in the template.

app/assets/stylesheets/application.scss
```diff
@@ -140,13 +137,13 @@ main {
...
-    padding: 0 0 10px;
-    border-bottom: 1px solid $border-colour;
+    @include govuk-font(16);
+    margin: govuk-spacing(2) 0 0 0;
+    padding: 0 0 govuk-spacing(2);
+    border-bottom: 1px solid $govuk-border-colour;
 
     strong {
-      font-weight: bold;
+      @include govuk-typography-weight-bold;
```
andysellick: Just curious - what does this mixin provide over the standard `font-weight: bold`?

app/assets/stylesheets/application-ie8.scss
```diff
@@ -1,6 +1,5 @@
 // BASE STYLESHEET FOR IE 8 COMPILER
 
-$is-ie: true;
-$ie-version: 8;
+$govuk-is-ie8: true;
```
andysellick: I haven't seen this before, is this documented somewhere in the Design System? I thought these were custom variables for this application.

app/assets/stylesheets/application.scss
kr8n3r: yes, it's to do with the MQ change

app/assets/stylesheets/application.scss
kr8n3r: actually we can just use 700 which is what above mixin pulls in (`$govuk-font-weight-bold`)

app/assets/stylesheets/application-ie8.scss
kr8n3r: with govuk-frontend you can build an ie8 stylesheet - docs here https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/supporting-internet-explorer-8.md

app/assets/stylesheets/_manuals-print.scss
kr8n3r: this stylesheet isn't importing govuk-frontend and the font still comes from govuk_template
will have a check again an do a selective import from govuk-frontend

app/views/layouts/application.html.erb
kr8n3r: oh didn't think to look there as well. thanks

kr8n3r: we can update `<!--[if gt IE 8]>` to `<!--[if lte IE 8]>`and serve those browsers that stylesheet and and any fixes in that stylesheet, plus import html5 shiv

we can't use the old 
```
$is-ie: true;
$ie-version: 8;
```
as govuk-frontend doesn't support it.

is there a better way while moving to govuk-frontend?

app/assets/stylesheets/application-ie8.scss
andysellick: Just so I understand, where did the previous `$is-ie` stuff come from? What makes this work?

Incidentally I just checked a few pages in this PR in IE8 and it actually looks better than the current live version 😲 

Screen Shot 2019-09-05 at 10 08 26

Screen Shot 2019-09-05 at 10 08 47

app/views/layouts/application.html.erb
andysellick: Updating the conditional is a good idea.

app/assets/stylesheets/application.scss
kr8n3r: good shout, i'll do that

app/views/layouts/application.html.erb
andysellick: Wait, hang on, wouldn't updating the conditional like that be exactly the same as removing the conditional entirely? It's saying if the browser is IE8 or below then use this stylesheet, and every other browser will use it anyway. Might as well just remove it.

app/assets/stylesheets/application-ie8.scss
kr8n3r: is-ie and ie-version come from frontend_toolkit. in govuk-frontend 'ie-version' was replaced with just `$govuk-is-ie8` as we dropped support of 6 and 7

app/views/layouts/application.html.erb
kr8n3r: not exactly. if you take away 6 and 7 stylesheets and don't change the ie8 conditiional, 6 & 7 get none of the application styles, just styles from govuk-template
Screen Shot 2019-09-05 at 12 06 29

app/assets/stylesheets/application.scss
36degrees: I'd definitely encourage people to use the mixin (or `govuk-font` with the `$weight` attribute) so that A) if we ever change the weight of the bold font (admittedly unlikely) it's easy to roll out but also B) if we ever introduce any additional complexity that depends on weight (like `font-size-adjust`, which was [dependent on font weight in frontend toolkit](https://github.com/alphagov/govuk_frontend_toolkit/blob/03204449dcbf28c2d48e7f6dd217f6abe9b3a518/stylesheets/_typography.scss#L31-L35) we can make sure those adjustments get applied auto-magically.

app/views/layouts/application.html.erb
kr8n3r: i've switched to `lte ie8` so 6 & 7 get the ie8 stylesheet

kr8n3r:
with the change legacy IE look

**IE6**
Screen Shot 2019-09-05 at 13 20 56
Screen Shot 2019-09-05 at 13 22 48

**IE7**
Screen Shot 2019-09-05 at 13 23 12
Screen Shot 2019-09-05 at 13 23 41

app/views/layouts/application.html.erb
kr8n3r: @andysellick see comment with screenshot from IE6&7

andysellick: That all looks great. I'm still a bit confused about how it's working, but don't worry.

PR #904, opened 2020-05-17
Run RuboCop as part of `bundle exec rake` defaults

issyl0:
- This is a prerequisite for being able to move more of our apps to
  GitHub Actions. We want one single step bundle exec rake that runs
  everything. This ensures parity between local dev and what CI runs
  (something that we don't always have now).

https://trello.com/c/5Gob6WUL/147-for-everyone-ensure-that-rubocop-runs-as-part-of-bundle-exec-rake-as-one-of-the-default-tasks

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -5,4 +5,5 @@ require File.expand_path("config/application", __dir__)
 
 ManualsFrontend::Application.load_tasks
 
-task default: ["jasmine:ci"]
+Rake::Task[:default].clear unless Rails.env.production?
+task default: [:lint, :spec, "jasmine:ci"]
--- a/lib/tasks/lint.rake
+++ b/lib/tasks/lint.rake
@@ -1,0 +1,4 @@
+desc "Linting"
+task lint: :environment do
+  sh "bundle exec rubocop --format clang"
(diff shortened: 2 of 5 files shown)


Rakefile
```diff
@@ -5,4 +5,4 @@ require File.expand_path("config/application", __dir__)
 
 ManualsFrontend::Application.load_tasks
 
-task default: ["jasmine:ci"]
+task default: [:lint, :spec, "jasmine:ci"]
```
benthorner, line 8: So this is a problem I have with the way RSpec integrates with Rake. We should:

- Apply the approach [here](https://github.com/alphagov/content-publisher/blob/master/Rakefile#L7) to other PRs around this
- Make a consistent decision about whether `lint` comes before/after `spec` (about 50/50) at the moment

issyl0: I think it should come before. But it is a bit annoying when you just want to see if a test passes on your WIP commit...

issyl0:
```suggestion
Rake::Task[:default].clear
task default: [:lint, :spec, "jasmine:ci"]
```

PR #950, opened 2020-07-28
Make app consistent with upgrades

cdccollins:
Making manuals-frontend consistent with the other apps that have been upgraded to Rails 6


Rhian:
Will need to re-review once conflicts in Gemfile and Gemfile.lock are addressed.

PR #952, opened 2020-08-04
Fix date ordering on manuals updates page

karlbaker02:
This PR fixes date ordering on the manuals updates page. Previously, HMRC manuals displayed updates from earliest to most recent, whereas other manuals displayed updates from most recent to the earliest.

The reason for this is because the change notes for HMRC manuals, as recorded in the content item, ranged from most recent to earliest, whereas for other manuals was from earliest to most recent. A `reverse` operation was then applied to these change notes in the code, which meant that HMRC manuals were then displayed in the wrong order.

To resolve this, instead sort by date descending to ensure ordering is consistent for different manuals.

Raised from [this Zendesk ticket](https://govuk.zendesk.com/agent/tickets/4181394).

### DIFF
--- a/spec/support/app_helpers.rb
+++ b/spec/support/app_helpers.rb
@@ -87,7 +87,7 @@ module AppHelpers
   end
 
   def expect_change_note(change_note, options)
-    within(".subsection-collection") do
+    within(".section-content") do
       expect(page).to have_link(options[:section_title], href: options[:section_href])
       expect(page).to have_content(change_note)
     end
--- a/spec/support/manual_helpers.rb
+++ b/spec/support/manual_helpers.rb
@@ -41,7 +41,6 @@ module ManualHelpers
         title: "This is the section on hot sauce",
         change_note: "Added section on hot sauce",
         published_at: "2016-06-10T10:15:36Z",
-        # published_at: "2014-06-25T09:17:27Z",
       },
       {
         base_path: "#{base_path}/guacamole",
--- a/spec/support/manual_helpers.rb
+++ b/spec/support/manual_helpers.rb
@@ -16,6 +16,42 @@ module ManualHelpers
     ).fetch("links")
   end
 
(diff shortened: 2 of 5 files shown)


spec/support/app_helpers.rb
```diff
@@ -87,7 +87,7 @@ def expect_page_to_be_affiliated_with_org(options)
   end
 
   def expect_change_note(change_note, options)
-    within(".subsection-collection") do
+    within(".section-content") do
```
karlbaker02, line 90: @leenagupte The tests were previously failing here because up until now there had only ever been one year used in the test data, which meant that there'd only ever be one `subsection-collection` class; by adding more test data across different years, it broke the Capybara `within` selector.

I've instead changed this to be the parent of all `subsection-collection`s, which is the `section-content` class. I'm not sure if it was previously scoped to a `subsection-collection` for any good reason, such as perhaps same-named links repeating across different years, but from the previous test data I don't believe that's the case.

spec/support/manual_helpers.rb
```diff
@@ -16,6 +16,42 @@ def example_links
...
+        title: "Added section on salsa",
+        change_note: "Added a section on salsa",
+        published_at: "2014-06-25T09:17:27Z",
+      },
+      {
+        base_path: "#{base_path}/this-is-the-section-on-hot-sauce",
+        title: "This is the section on hot sauce",
+        change_note: "Added section on hot sauce",
+        published_at: "2016-06-10T10:15:36Z",
+        # published_at: "2014-06-25T09:17:27Z",
```
leenagupte, line 44:
```suggestion
```

leenagupte: Should probably remove this.

karlbaker02: Thanks, didn't see this. Will remove 👍

PR #969, opened 2020-09-15
Move phase banner and contextual breadcrumbs

danacotoran:
## What
Moved phase banner and contextual breadcrumbs outside of main in manuals
frontend. 

## Why
`<main>` should only contain elements that are unique to the
document, which the breadcrumbs and phase banner are not.

## Visual changes
This should not result in any visual changes

https://trello.com/c/UbK2Cher


owenatgov (approved):
👌

PR #982, opened 2020-10-09
Fix HMRC manuals header WCAG fail

danacotoran:
## Problem

On HMRC manual pages such as [this one](https://www.gov.uk/hmrc-internal-manuals/paye-manual/paye70201), the turquoise background and white text of the content header don't meet the minimum contrast requirements for regular sized text.

The adjacent colours against the search button when it is focused don't meet the minimum colour contrast requirements.

This fails WCAG SC 1.4.3 (content header) and 1.4.11 (search button).

## Solution
Use websafe version of the HMRC brand colour, which is conducive to better contrast, to ensure the header text is readable.

### Before (top) vs after (bottom)
Screenshot 2020-10-09 at 18 00 54

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -44,7 +44,7 @@ main {
     background: $govuk-brand-colour;
 
     &.hmrc {
-      background: govuk-organisation-colour("hm-revenue-customs", $websafe: false);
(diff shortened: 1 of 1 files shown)


injms (approved):
🎨 👍

PR #995, opened 2020-10-28
Update govspeak component usage

maxgds:
### What
Updates the way the Govspeak component to not use the content parameter.

### Why
The Govspeak component should not be used with the content parameter as that will be deprecated; it should use a do block wrapped around sanitized markup instead.

See alphagov/govuk_publishing_components#1632 for more info.

### Visual differences
None.


injms (approved):
👍

PR #997, opened 2020-10-28
Revert calls to govspeak component

maxgds:
This reverts https://github.com/alphagov/manuals-frontend/pull/995 as the added sanitize call is causing table markup to fail. Initially overlooked as checked pages didn't contain tables. Will look at safer implementation after reverting.


benthorner (approved):
👍

PR #1003, opened 2020-11-04
Fix duplicated body text in manual section

theseanything:
Removing a extra render of the body text in the manual section.


owenatgov (approved):
💪

PR #1060, opened 2021-02-05
Replace manuals accordion with gem accordion

owenatgov:
## What
Replace the custom manuals accordion (collapsible collection) with the [components gem accordion](https://components.publishing.service.gov.uk/component-guide/accordion). This includes:

- Moving the processing of the markup for individual manuals from the js to the document model
- Updating the updates view
- Removing evidence of the previous accordion implementation. This includes all of jasmine as there are now no js modules within this repo to test and therefore no point in including the test suite

## Why
This is part of work within the govuk accessibility team to ensure that accordions are consistent across gov.uk. Manuals frontend has been using a bespoke accordion implementation which bolts on accordion functionality on top of govspeak content from manuals publisher. This creates risk because any updates to our accordion can't reliably permeate to manuals and could create hidden legacy code and accessibility issues. This change creates a relationship between manuals and our components gem, as is the case with our other frontend apps, to minimise risk and increase consistency between manuals and the rest of gov.uk.

The team are conscious that this is, in a sense, perpetuating the hack and the ideal would be to change the publishing workflow to account for accordion content, thus mitigating the need for any sort of preprocessing of the content item. We don't have the resource or knowledge in the team to action this, however [an issue has been raised](https://github.com/alphagov/manuals-publisher/issues/1713) addressing this need.

[Card](https://trello.com/c/JNcidGoj/609-technical-spike-into-how-we-might-update-the-manual-accordion-component-and-implement-the-fix)

### Pages tested against
- [Manuals page](https://gov.uk/guidance/working-safely-during-coronavirus-covid-19/construction-and-other-outdoor-work)
- [Manuals page with advanced accordion markup](https://www.gov.uk/guidance/mot-inspection-manual-for-private-passenger-and-light-commercial-vehicles/3-visibility)
- [HMRC manuals page that doesn't use accordions but still relies on the same document model as those using accordions to render content](https://gov.uk/hmrc-internal-manuals/employment-income-manual/eim00500)
- [Manual updates page](https://gov.uk/guidance/working-safely-during-coronavirus-covid-19/updates)

### Before
Screenshot 2021-02-05 at 14 55 31
Screenshot 2021-02-05 at 14 56 43

### After
Screenshot 2021-02-12 at 10 41 46
Screenshot 2021-02-12 at 10 42 14


app/views/manuals/_manual_section.html.erb
```diff
@@ -9,7 +9,7 @@
       <% if presented_document.body.present? %>
         <div class='body-content-wrapper'>
           <%= render 'govuk_publishing_components/components/govspeak', {} do %>
-            <%= raw(presented_document.body) %>
+            <%= sanitize(presented_document.body[:intro]) %>
```
maxgds, line 12: I might be being overly cautious and this may be fine, but we need to be careful of this as it can break tables etc: check this url against your version: https://www.gov.uk/guidance/mot-inspection-manual-for-private-passenger-and-light-commercial-vehicles/3-visibility

owenatgov: Aha, this does indeed break tables 😬  I'll investigate

owenatgov:
### Note on tests
End to end test are currently failing. This is because this PR introduced a test failure in the publishing-e2e-tests repo due to the significant change in markup. [I've created a separate PR to handle this](https://github.com/alphagov/publishing-e2e-tests/pull/413). As per the e2e contribution guidance, once this has been reviewed and approved I'll set this branch to test against my e2e branch, merge my changes, deploy and then merge my e2e PR.

app/views/manuals/_manual_section.html.erb
maxgds: I think it may have been manuals where I looked into this before - we had to roll back a release where we'd broken tables on live, but when I looked into what we needed to allow on the sanitize list it was pretty much everything - probably best to leave it as raw for now.

owenatgov: Makes sense, was a good spot, cheers 👍  As it happens this was mainly put in to fix a test, however upon revision I think I've come up with a solution that uses nokigiri properly and is much more readable.

ChrisBAshton (commented):
LGTM - but made some code quality suggestions below.

I'd also amend the last commit to include the context you provided on Slack around why you're removing Jasmine 👍

app/views/manuals/_manual_section.html.erb
```diff
@@ -9,7 +9,7 @@
       <% if presented_document.body.present? %>
         <div class='body-content-wrapper'>
           <%= render 'govuk_publishing_components/components/govspeak', {} do %>
-            <%= raw(presented_document.body) %>
+            <%= raw(presented_document.body[:intro]) %>
```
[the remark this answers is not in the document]
owenatgov: I was initially worried about passing the entire body plus the broken down body and intro in a single object. On balance though, would this have any noticeable user ramifications? Your reasoning makes sense and would certainly make it easier to manage.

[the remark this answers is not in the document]
ChrisBAshton: What worries you about that? Are you worried it may not be performant enough on the server? We can utilise caching if that helps, e.g. this pseudocode:

```ruby
def intro
  body.strip_out_intro
end

def main
  body.strip_out_main
end

def body
  # the following line is the equivalent to `@body = @body || some_expensive_operation`
  @body ||= some_expensive_operation
end
```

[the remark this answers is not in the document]
owenatgov: I think I assumed there would be a performance hit yeah but I think this was just me making very rough guesses of how ruby worked. Your other point on breaking this apart, plus the pseudo code above mitigates this. I'll sort it out. Cheers!

app/models/document.rb
```diff
@@ -36,7 +36,41 @@ def summary
   end
 
   def body
-    details["body"] && details["body"].html_safe
+    if details["body"] && details["body"].html_safe
+      processed_body = []
+      document = Nokogiri::HTML::DocumentFragment.parse(details["body"])
+      intro_content = document.clone
+
+      # Get first block of content
```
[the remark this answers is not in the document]
owenatgov: I've tried to describe what this does in a bit more detail. I can add some more details to the intro method or main if you think it's required.

app/views/manuals/_manual_section.html.erb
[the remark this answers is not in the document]
owenatgov: I've broken the methods in this down into `body`, `main` and `intro` where the latter 2 methods consume `body`. It's not the exact implementation you specified above but I think it's a lot neater than it was. Let me know if this is a problem and I can re-amend.

app/models/document.rb
```diff
@@ -39,6 +39,49 @@ def body
...
+  def main
+    return nil unless body
+
+    document = Nokogiri::HTML::DocumentFragment.parse(body)
+    main = []
+    
+    # Identifies all headings of the level specified in collapse_path and creates an array of objects from the heading and its proceeding content up to the next heading
+    # This is so that it can be consumed by accordion components in the template
+    # See _manual_section.html.erb for how this is being rendered
+    document.css("h#{collapse_depth}").each do |heading|
```
ChrisBAshton, line 63: This could be a map:

```ruby
main = document.css("h#{collapse_depth}").map do |heading|
  # ...
  {
    heading: ...
    content: ...
  }
end
```

And rather than defining `main` and then returning it, we can just return the map directly:

```diff
- main = document.css("h#{collapse_depth}").map do |heading|
+ document.css("h#{collapse_depth}").map do |heading|
```

app/models/document.rb
```diff
@@ -39,6 +39,49 @@ def body
...
+
+    intro
+  end
+
+  def main
+    return nil unless body
+
+    document = Nokogiri::HTML::DocumentFragment.parse(body)
+    main = []
+
```
ChrisBAshton, line 59: Linter is complaining about extra whitespace:

```suggestion
```

ChrisBAshton (approved):
LGTM 👍 thanks for making the changes! One or two non-blocking comments below:

app/models/document.rb
owenatgov: Done!

PR #1071, opened 2021-02-16
Render newlines in manual update notes

bilbof:
Change notes (what we render on /guidance/`<manual>`/updates) can have newline characters. We were stripping them, which makes it hard to read (and worse when printing the page).

Using [simple_format](https://api.rubyonrails.org/classes/ActionView/Helpers/TextHelper.html#method-i-simple_format) renders the newlines instead, making the page easier to read.

simple_format also sanitizes the change notes.


beccapearce (approved):
Ooh nice spot!

PR #1072, opened 2021-02-17
Add anchor navigation feature flag to accordions

owenatgov:
## What
Adds an id attribute to the `heading` object on accordion rendering, as well as the `anchor_navigation` feature flag to allow anchor navigation on manuals page accordions.

## Why
Part of a fix of the previous feature allowing manuals to link to individual accordion sections on other manuals.

requires [this PR on the components gem](https://github.com/alphagov/govuk_publishing_components/pull/1937) to be released before his can be merged and deployed.

No visual changes.

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -73,6 +73,7 @@ class Document
       {
         heading: {
           text: heading.text,
+          id: heading[:id],
         },
         content: content.join,
       }
--- a/app/views/manuals/_manual_section.html.erb
+++ b/app/views/manuals/_manual_section.html.erb
@@ -28,6 +28,7 @@
         end %>
 
         <%= render "govuk_publishing_components/components/accordion", {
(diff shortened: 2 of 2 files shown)


app/models/document.rb
```diff
@@ -73,6 +73,7 @@ def main
       {
         heading: {
           text: heading.text,
+          id: heading[:id],
```
chris-gds, line 76: I'm testing the links on this page

`/how-to-publish-on-gov-uk/creating-and-updating-pages`

There is a link to `#attachments` however the actual ID is `adding-attachments`

From viewing the workflow I'm thinking the IDs are done manually vs programmatically. Investigating a little more

owenatgov: If you test with the latest version of the gem PR this should be working as expected

chris-gds (approved):

PR #1078, opened 2021-03-04
Return 406 Not Acceptable when 'Accepts' header requests non-HTML

ChrisBAshton:
We've had a number of Sentry issues to do with requesting JSON or
JS versions of Manuals Frontend pages:
- https://sentry.io/organizations/govuk/issues/2152998596/?project=202235&query=is%3Aunresolved
- https://sentry.io/organizations/govuk/issues/2161602206/?project=202235&query=is%3Aunresolved

The correct thing to do is to respond with a [406](https://httpstatuses.com/406),
and not to log inactionable Sentry issues.

I've taken inspiration from Content Publisher for the request
testing, and the underlying fix is the same approach as in
government-frontend:
https://github.com/alphagov/government-frontend/blob/ce26ae94d66d9ce92ac37028854211741ef44b8b/app/controllers/content_items_controller.rb#L6-L7

We will also need to do some work to ensure that 406 responses
are cached in Fastly.

Trello: https://trello.com/c/XNQhMGRt/2391-3-make-manuals-frontend-respond-to-bad-requests-with-406-status

### DIFF
--- a/spec/requests/request_spec.rb
+++ b/spec/requests/request_spec.rb
@@ -1,0 +1,47 @@
+require "rails_helper"
+
+RSpec.describe "Manuals" do
+  describe "GET ManualsController#index" do
+    it "returns successfully" do
+      slug = "/guidance/my-manual-about-burritos"
+      stub_fake_manual(base_path: slug)
+      get slug
+
+      expect(response).to have_http_status(:ok)
+    end
+  end
+
+  describe "GET ManualsController#index with unacceptable 'Accept' header" do
+    it "returns 406 Not Acceptable" do
+      slug = "/guidance/my-manual-about-burritos"
+      stub_fake_manual(base_path: slug)
+      get slug, params: {}, headers: { "Accept" => "application/json" }
+
+      expect(response).to have_http_status(:not_acceptable)
+    end
+  end
+
+  describe "GET ManualsController#show" do
+    it "returns successfully" do
+      slug = "/guidance/my-manual-about-burritos"
+      stub_fake_manual(base_path: slug)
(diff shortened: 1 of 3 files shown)


spec/requests/request_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      section = "/guidance/my-manual-about-burritos/section"
+      stub_fake_manual(base_path: section)
+      get section
+
+      expect(response).to have_http_status(:ok)
+    end
+  end
+
+  describe "GET ManualsController#show" do
+    it "returns successfully" do
```
brucebolt, line 37:
```suggestion
    it "returns 406 Not Acceptable" do
```

spec/requests/request_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      stub_fake_manual(base_path: slug)
+      section = "/guidance/my-manual-about-burritos/section"
+      stub_fake_manual(base_path: section)
+      get section
+
+      expect(response).to have_http_status(:ok)
+    end
+  end
+
+  describe "GET ManualsController#show" do
```
brucebolt, line 36:
```suggestion
  describe "GET ManualsController#show with unacceptable 'Accept' header" do
```

brucebolt (changes requested):
Looks good to me, except one of the tests needs two small changes in the descriptions

spec/requests/request_spec.rb
ChrisBAshton: Oops 🤦

brucebolt (approved):

PR #1082, opened 2021-03-09
ManualSections that are"visually_expanded" render without accordions

edwardkerry:
Manual sections that are published as `"visually_expanded": true`
will render as flat text  without the "accordion" govuk_publishing_component.
https://components.publishing.service.gov.uk/component-guide/accordion

The default publishing option is `"visually_expanded": false`. If the
`visually_expanded` attribute is not set the section will also continue
to display as an accordion.

Also removes an unused `collapsible?` method from the Section model.

#### Default `"visually_expanded": false` or absent
Screenshot 2021-03-09 at 10 53 08

------

#### Specified `"visually_expanded": true`
Screenshot 2021-03-09 at 10 54 43

[govuk-content-schemas](https://github.com/alphagov/govuk-content-schemas/pull/1051)
[trello](https://trello.com/c/wgEKW9Ze/2398-5-allow-manuals-publisher-to-set-accordions-to-open)

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -39,6 +39,10 @@ class Document
     details["body"] && details["body"].html_safe
   end
 
+  def visually_expanded?
+    details.fetch("visually_expanded", false)
+  end
+
   def intro
     return nil unless body
 
--- a/app/models/section.rb
+++ b/app/models/section.rb
@@ -19,10 +19,6 @@ class Section
     section["base_path"]
   end
 
-  def collapsible?
-    body.present?
-  end
-
   def summary
     section["description"]
   end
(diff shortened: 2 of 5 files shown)


1pretz1 (approved):
Looks great Ed. Thanks for providing screenshots!

The only niggle I have is I wish the app used factories (or fixtures) as most of the stub methods are quite heavy with config, but this isn't the concern of this PR! 😄

callumknights (approved):
I was thinking the same about that response config Pete, so thanks for that little bit of information!

thomasleese (approved):

benjamineskola (approved):

PR #1083, opened 2021-03-09
Prevent manual_section title from appearing twice

edwardkerry:
A change to add the breadcrumb to the document heading for
manual_section view has caused the title to sometimes appear twice.

The document_heading is intended to present the `section_id - title`.
The intention is that the `breadcrumb` method should provide the `section_id` - however
if this is not present it nstead defaults to `title`.

https://github.com/alphagov/manuals-frontend/blob/63c17fff271e85fcf9f6196963e0d4135214c29f/app/models/document.rb#L26

This change prevents the title being added to the document_heading if it
is already present i.e: the `breadcrumb` method was unable to return a `section_id`.

### DIFF
--- a/app/views/manuals/_manual_section.html.erb
+++ b/app/views/manuals/_manual_section.html.erb
@@ -1,12 +1,12 @@
 <%
   document_heading = []
-  document_heading << "#{presented_document.breadcrumb} - " if presented_document.breadcrumb.present?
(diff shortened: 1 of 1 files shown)


injms (approved):

injms:
Sorry I missed this. Can we add an example of a page where this happened in the readme for future checking?

PR #1085, opened 2021-03-09
Apply two thirds width to accordion sections in manuals and manual updates

owenatgov:
## What
Sets a two thirds limit on desktop for manual body content where accordions are rendered.

Additionally, this PR does some spring cleaning adjacent to this view:
- Adds a `govuk-width-container` class to the `id="wrapper"` content container div
- Replaces the lead paragraph on manuals with the [lead paragraph component](https://components.publishing.service.gov.uk/component-guide/lead_paragraph)

## Why
This is in response to feedback from department publishers that accordions are now very long which isn't a good experience for magnification users.

[Card](https://trello.com/c/wfX1mzXP/675-fix-accordion-column-text-width)

### Pages developed against
- https://www.gov.uk/guidance/how-to-publish-on-gov-uk/creating-and-updating-pages
- https://www.gov.uk/guidance/working-safely-during-coronavirus-covid-19/updates
- https://www.gov.uk/hmrc-internal-manuals/holding-and-movements-assurance-guidance/hmag60320

### Before - manual
Screenshot 2021-03-09 at 15 12 22

### Before - updates
Screenshot 2021-03-09 at 15 05 27

### After - manual
Screenshot 2021-03-09 at 15 16 13

### After - updates
Screenshot 2021-03-09 at 18 10 55

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -13,6 +13,7 @@ $govuk-use-legacy-palette: false;
 @import 'govuk_publishing_components/components/hint';
 @import 'govuk_publishing_components/components/input';
 @import 'govuk_publishing_components/components/label';
+@import 'govuk_publishing_components/components/lead-paragraph';
 @import 'govuk_publishing_components/components/metadata';
 @import 'govuk_publishing_components/components/phase-banner';
 @import 'govuk_publishing_components/components/previous-and-next-navigation';
@@ -136,30 +137,12 @@ main {
   }
 
   .manual-body {
-    margin: govuk-spacing(3) 0 90px;
-
     @include govuk-media-query($from: tablet) {
       margin-top: govuk-spacing(6);
     }
 
     .section-title {
       @extend %govuk-heading-m;
-    }
-
-    .summary {
-      @include govuk-font(24);
-      margin-bottom: govuk-spacing(3);
-
-      @include govuk-media-query($from: tablet) {
-        max-width: 66.6666%;
-        margin-bottom: govuk-spacing(6);
-      }
-    }
-
-    .body-content-wrapper {
-      @include govuk-media-query($from: tablet) {
-        width: 66.6666%;
-      }
     }
 
     .footnotes {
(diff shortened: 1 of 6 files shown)


app/views/manuals/show.html.erb
```diff
@@ -23,10 +23,15 @@
...
-<%=
-  render(
-    'manual_section',
-    presented_manual: presented_manual,
-    presented_document: presented_document
-  )
-%>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds-from-desktop">
```
maxgds, line 28: Based on looking at a manuals page, I wonder if this should be from-tablet rather than from-desktop? eg https://manuals-fron-two-thirds-jnnriy.herokuapp.com/hmrc-internal-manuals/holding-and-movements-assurance-guidance/hmag60320

I suggest going to the national archives or archive.org to confirm what it used to do and matching if you didn't already (apologies if you did!).

owenatgov: It appears that govspeak is applying it's own width wrapper for HMRC manuals 😬  I'll amend this to not include HMRCs at the very least and investigate what the original implementation was.

owenatgov: Ah, that's not quite right. There's a custom width wrapper in the markup itself for HMRC manuals. I'll just scoop that out then, easy.

app/assets/stylesheets/application.scss
```diff
@@ -13,6 +13,7 @@ $govuk-use-legacy-palette: false;
 @import 'govuk_publishing_components/components/hint';
 @import 'govuk_publishing_components/components/input';
 @import 'govuk_publishing_components/components/label';
+@import 'govuk_publishing_components/components/lead-paragraph';
```
chris-gds, line 16: This is a good idea 👏🏼

chris-gds:
All functions and looks good, prefer the approach of using the grid rather than restricting the inner content. 🙌🏼.

chris-gds (approved):

PR #1113, opened 2021-04-27
Handle gone status

1pretz1:
Sentry error: https://sentry.io/organizations/govuk/issues/1752513075/?environment=production&project=202235&query=is%3Aignored

Previously the code would raise an error if `document_from_repository`
returned a document that had a status of 'Gone'. This was raised as
[HTTPGone][1] from gds-api-adapters as a result of receiving a 410 from
content-store.

Now 'Gone' pages and their corresponding exception are rescued and
handled by returning a 410 status code which in turn renders a [generic
410 error page][2].

I've also refactored the error handling logic as part of this work
which:

* moves all this behaviour to the manuals controller
* follows the pattern in [government-frontend][3]

[1]: https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/exceptions.rb#L90-L91
[2]: https://github.com/alphagov/static/blob/master/app/views/root/410.html.erb
[3]: https://github.com/alphagov/government-frontend/blob/e5c0c97dc614b9513b82e106074e9293826498a4/app/controllers/content_items_controller.rb#L2-L10

### DIFF
--- a/app/controllers/manuals_controller.rb
+++ b/app/controllers/manuals_controller.rb
@@ -1,6 +1,11 @@
 class ManualsController < ApplicationController
-  before_action :ensure_manual_is_found
-  before_action :ensure_document_is_found, only: :show
+  rescue_from GdsApi::HTTPForbidden, with: :error_403
+  rescue_from GdsApi::HTTPNotFound, with: :error_404
+  rescue_from GdsApi::HTTPGone, with: :error_410
+  rescue_from ActionView::MissingTemplate, with: :error_406
+  rescue_from ActionController::UnknownFormat, with: :error_406
+
+  before_action :follow_redirect_if_document_redirected, only: :show
 
   def index
     set_expiry(content_store_manual)
@@ -34,47 +39,11 @@ class ManualsController < ApplicationController
 private
 
   def presented_manual
-    @presented_manual ||=
-      ManualPresenter.new(manual: manual, view_context: view_context)
-  end
-
-  def ensure_manual_is_found
-    if content_store_manual.nil?
-      error_not_found
-    elsif content_store_manual["format"] == "gone"
-      error_gone
-    end
-  end
-
-  def ensure_document_is_found
-    if document_from_repository.blank?
-      error_not_found
-    elsif document_from_repository["document_type"] == "redirect"
-      error_redirect
-    end
-  end
-
-  def error_redirect
-    destination, status_code = GdsApi::ContentStore.redirect_for_path(
-      document_from_repository, request.path, request.query_string
-    )
-    redirect_to destination, status: status_code
-  end
-
-  def error_not_found
-    head :not_found
-  end
-
-  def error_gone
-    head :gone
+    @presented_manual ||= ManualPresenter.new(manual: manual, view_context: view_context)
   end
 
   def content_store_manual
-    @content_store_manual ||= begin
-      GdsApi.content_store.content_item(manual_base_path)
-                              rescue GdsApi::ContentStore::ItemNotFound
-                                nil
-    end
+    @content_store_manual ||= GdsApi.content_store.content_item(manual_base_path)
   end
 
   def manual
@@ -90,8 +59,7 @@ private
   end
 
   def document_from_repository
-    @document_from_repository ||=
-      DocumentRepository.new.fetch(document_base_path)
+    @document_from_repository ||= DocumentRepository.new.fetch(document_base_path)
   end
 
   def document_base_path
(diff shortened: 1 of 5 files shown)


app/controllers/manuals_controller.rb
```diff
@@ -34,47 +39,11 @@ def updates
 private
 
   def presented_manual
-    @presented_manual ||=
-      ManualPresenter.new(manual: manual, view_context: view_context)
-  end
-
-  def ensure_manual_is_found
-    if content_store_manual.nil?
```
ChrisBAshton, line 42: Do we still need to keep the `.nil?` test?

app/controllers/manuals_controller.rb
```diff
@@ -1,6 +1,11 @@
 class ManualsController < ApplicationController
-  before_action :ensure_manual_is_found
-  before_action :ensure_document_is_found, only: :show
```
ChrisBAshton, line 3: Are we losing some functionality here?

app/controllers/manuals_controller.rb
```diff
@@ -107,4 +75,29 @@ def set_expiry(content_item)
     cache_private = content_item.cache_control.private?
     expires_in(max_age, public: !cache_private)
   end
+
+  def follow_redirect_if_document_redirected
```
ChrisBAshton, line 79: Is this new functionality? If so, could it be its own separate commit?

spec/features/viewing_a_manual_spec.rb
```diff
@@ -175,22 +175,23 @@
     expect(page.status_code).to eq(404)
   end
 
-  scenario "visiting a withdrawn section" do
```
ChrisBAshton, line 178: Are we no longer testing withdrawn documents?

ChrisBAshton (commented):
Thanks for jumping on this, Peter. The code looks great, I just have a few questions about functionality that may have slipped through the cracks (if it's covered by the new code, it's not obvious to me).

spec/features/viewing_a_manual_spec.rb
1pretz1: I think think the terminology of 'withdrawn' is confused with documents that are actually 'gone' so I updated the wording. I did however forget to remove the [stub method](https://github.com/alphagov/manuals-frontend/pull/1113/files#diff-a8c2a229533b0ffd4bf783bbf1a0ae45d28016930dea8062f410d9d112dbb52bL326-L334) that this test used which as you can see from the format is actually a 'gone' document.

app/controllers/manuals_controller.rb
1pretz1: No new functionality, just a slightly different name. The old logic resided in a [different method](https://github.com/alphagov/manuals-frontend/pull/1113/files#diff-6b5ff7cba3fc3659ddf2782828d61b92542869a958e4970b0d650d108ba881c0L52-L54).

app/controllers/manuals_controller.rb
1pretz1: I don't think so as the functionality these callbacks provided are now represented by the `follow_redirect_if_document_redirected` callback and the [rescue blocks](https://github.com/alphagov/manuals-frontend/pull/1113/files#diff-6b5ff7cba3fc3659ddf2782828d61b92542869a958e4970b0d650d108ba881c0R2-R4). These were previously handled by [rescuing](https://github.com/alphagov/manuals-frontend/pull/1113/files#diff-6b5ff7cba3fc3659ddf2782828d61b92542869a958e4970b0d650d108ba881c0L75-L77) at a different point and comparing [nils](https://github.com/alphagov/manuals-frontend/pull/1113/files#diff-6b5ff7cba3fc3659ddf2782828d61b92542869a958e4970b0d650d108ba881c0L42). 

The rescue classes are different than what was previously implemented but they mean the same thing. E.g in [gds-api-adapters](https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/content_store.rb#L7) the `GdsApi::ContentStore::ItemNotFound` error class inherits from `GdsApi::HTTPNotFound`.

app/controllers/manuals_controller.rb
1pretz1: I don't think so, see the comment below.

1pretz1:
Thanks for the review @ChrisBAshton! I should've answered all your questions but let me know if you need clarity on any of the points above. An extra stub method has also been removed as a result so success. 🎉

ChrisBAshton (approved):
Thanks for confirming, Peter. All looks good then! 👍

PR #1269, opened 2022-02-24
Remove jQuery from manuals-frontend

andysellick:

## What
Remove jQuery from this application.

- only jQuery was this document.ready equivalent for primaryLinks
- replacing this is possible but slightly complicated by browser compatibility issues
- so maybe we could remove it entirely
- we've already removed document.ready successfully elsewhere on GOV.UK without a direct replacement, and testing this here it seems to work as well
- primaryLinks is still initialised and runs successfully

## Why
We're removing jQuery from GOV.UK.

## Visual changes
None.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -5,7 +5,6 @@
 //= require govuk_publishing_components/components/govspeak
 //= require govuk_publishing_components/components/print-link
 
-jQuery(function ($) {
-  'use strict'
+if (GOVUK.primaryLinks) {
   GOVUK.primaryLinks.init('.primary-item')
-})
+}


alex-ju:
I think we still have this script running on a few pages (via govspeak, which is why we can't find the markup) – more details on this here https://github.com/alphagov/govuk_publishing_components/issues/2364

andysellick:
Ah, I forgot about govspeak 🤦  Thanks @alex-ju, I'll have another think about this.

andysellick:
@alex-ju have tried an alternative approach with this...

alex-ju (approved):

alex-ju:
> @alex-ju have tried an alternative approach with this. I'm wondering now if I should wrap it in a check first, is this sensible?
> 
> ```
> if (GOVUK.primaryLinks) {
>   GOVUK.primaryLinks.init('.primary-item')
> }
> ```

Wrapping it would indeed make it more resilient to an potential accidental removal of the primaryLinks script.

